PR #210, opened 2024-01-22
chore: Remove private repo thingy

gitferry:


Dockerfile
```diff
@@ -14,7 +14,6 @@ RUN apk add --no-cache --update openssh git make build-base linux-headers libc-d
 
 RUN mkdir -p /root/.ssh && ssh-keyscan github.com >> /root/.ssh/known_hosts
```
SebastianElvis, line 15: Can this be gone as well? also L16 and L23

gitferry: I think these lines are irrelevant as the ssh keys are from the container instead of our local machines

SebastianElvis: Given that everything is public now, do we still need these ssh thingy?

gitferry: Did some investigation and you are right. Removed them. Thanks!

SebastianElvis (approved):
finally we get rid of the git ssh things!

Issue #145, opened 2023-11-22
EOTS manager still asks passphrase from terminal in `file` KeyringBackend


gitferry:
It happens if the keyring backend is `file`

gitferry:
The issue is that if the `eotsd` is running in a terminal, when creating a key, it will [ask passphrase from a prompt](https://github.com/cosmos/cosmos-sdk/blob/6d10c507842e220aa43ad821e41272b3dd54a453/client/input/input.go#L22). Not sure how to bypass it yet

PR #214, opened 2024-01-23
fix(doc): Inconsistency of keyringbackend in doc

gitferry:
As we have keyring backend of `test` as the default value. This should be consistent with the docs. Also, due to issue #145, `file` keyring backend is not well supported (user needs to type the passphrase prompted from the eotsd's terminal).


vitsalis (approved):
Update the docs website as well?

PR #225, opened 2024-01-31
fix: Only clean up blocks buffer up to the fast-synced height

vitsalis:
In the following scenario:

1. Fast sync starts
2. A new block is retrieved by the poller
3. Fast sync ends

Step 2, adds the new block to the poller blocks channel (which is not known to the fast sync process). However, at step 3, the fast sync process cleans up the block cache, leading to the new block getting lost.


gitferry (approved):
Thanks for the fix!

Issue #227, opened 2024-01-31
bug: `LastProcessedHeight` leads to blocks never being voted if previously skipped


vitsalis:
There are two scenarios:

1. During normal operation, a block is skipped for some reason (e.g. due to a bug like the one on #225 ) and the next one is voted (e.g. heights `N` and `N+1`). The last processed height is set to `N+1`, leading to the block at height `N` never being voted for.
2. During fast sync, some blocks might be skipped due to a call to [shouldSubmitFinalitySignature](https://github.com/babylonchain/finality-provider/blob/8473c52206998efe5f542ca772d8694d625b6764/finality-provider/service/fastsync.go#L52), which might return `false` even for cases where a signature should be submitted to the future (e.g. no public randomness available right now). The end of the fast sync [sets LastProcessedHeight](https://github.com/babylonchain/finality-provider/blob/8473c52206998efe5f542ca772d8694d625b6764/finality-provider/service/fastsync.go#L89) to the latest height of the sync, which means that those blocks won't ever be revisited. If it was set to `syncedHeight` we wouldn't have this issue.

In general, the first scenario can be considered as a failsafe for recovering in the case of bugs. The second scenario is a bug and should be resolved (also related to #226 ).

gitferry:
addressed by #228

Issue #226, opened 2024-01-31
bug: shouldSubmitFinalitySignature does not check whether randomness is committed


vitsalis:
[shouldSubmitFinalitySignature](https://github.com/babylonchain/finality-provider/blob/8473c52206998efe5f542ca772d8694d625b6764/finality-provider/service/fp_instance.go#L397) does not check whether public randomness has been committed.

Ideally, the following addition would suffice
```
       lastCommitedHeight, err := fp.GetLastCommittedHeight()
       if err != nil {
               return false, err
       }

       if lastCommitedHeight < b.Height {
               fp.logger.Debug(
                       "the finality provider has not committed public randomness for the height",
                       zap.String("pk", fp.GetBtcPkHex()),
                       zap.Uint64("block_height", b.Height),
                       zap.Uint64("last_committed_height", lastCommitedHeight),
               )
               return false, nil
       }
```

Tried adding it but some of the tests fail and didn't have the bandwidth to resolve.

gitferry:
addressed by #228

PR #228, opened 2024-02-01
fix: `LastProcessedHeight` should be set ensuring public randomness committed

gitferry:
This PR fixed the bug documented in #226 and #227 by:
1. only setting `LastProcessedHeight` within `shouldSubmitFinalitySignature` after the check of public randomness existence, and 
2. adding the check of the last committed height within `shouldSubmitFinalitySignature`.

TODO:

- [ ] add e2e test to cover the case when the finality provider restarts with no available public randomness


finality-provider/service/fastsync_test.go
```diff
@@ -19,27 +20,21 @@ func FuzzFastSync(f *testing.F) {
...
-		mockClientController.EXPECT().
-			CommitPubRandList(fpIns.MustGetBtcPk(), startingBlock.Height+1, gomock.Any(), gomock.Any()).
-			Return(&types.TxResponse{TxHash: expectedTxHash}, nil).AnyTimes()
-		mockClientController.EXPECT().QueryFinalityProviderVotingPower(storeFp.MustGetBTCPK(), gomock.Any()).
+		mockClientController.EXPECT().QueryFinalityProviderVotingPower(fpIns.MustGetBtcPk(), gomock.Any()).
 			Return(uint64(1), nil).AnyTimes()
-		res, err := fpIns.CommitPubRand(startingBlock)
-		require.NoError(t, err)
-		require.Equal(t, expectedTxHash, res.TxHash)
+		lastCommittedHeight := randomStartingHeight + 25
```
vitsalis, line 30:
```suggestion
		lastCommittedHeight := randomStartingHeight + testutil.TestNumPubRand
```

finality-provider/service/fastsync_test.go
```diff
@@ -19,27 +20,21 @@ func FuzzFastSync(f *testing.F) {
...
+		lastCommittedPubRandMap := make(map[uint64]*btcec.FieldVal)
+		lastCommittedPubRandMap[lastCommittedHeight] = testutil.GenPublicRand(r, t).ToFieldVal()
+		mockClientController.EXPECT().QueryLastCommittedPublicRand(gomock.Any(), uint64(1)).Return(lastCommittedPubRandMap, nil).AnyTimes()
```
vitsalis, line 34: Recommend we also update the test to make sure that no blocks that don't have public randomness are processed and `LastProcessedHeight` is set appropriately.

finality-provider/service/fastsync.go
```diff
@@ -86,11 +86,9 @@ func (fp *FinalityProviderInstance) FastSync(startHeight, endHeight uint64) (*Fa
 		)
 	}
 
-	fp.MustSetLastProcessedHeight(endHeight)
```
vitsalis, line 89: Could this removal lead to a scenario in which we continuously try to vote for a block for which we don't have voting power?

finality-provider/service/fp_instance.go
```diff
@@ -405,18 +389,46 @@ func (fp *FinalityProviderInstance) shouldSubmitFinalitySignature(b *types.Block
...
 	if power == 0 {
 		fp.logger.Debug(
 			"the finality-provider does not have voting power, skip voting",
 			zap.String("pk", fp.GetBtcPkHex()),
 			zap.Uint64("block_height", b.Height),
 		)
+		// we need to set last processed height here
+		// as this ensures that the corresponding
+		// public randomness has already been committed
+		fp.MustSetLastProcessedHeight(b.Height)
```
vitsalis, line 431: Not in favour of adding this here. This is a query method, i.e. a method that responds to a question. It should not modify state.

finality-provider/service/fastsync.go
gitferry: I think we need to update this value per block so we don't need to set it here

finality-provider/service/fastsync_test.go
gitferry: added a case where the randomness is not sufficient

finality-provider/service/fp_instance.go
```diff
@@ -184,36 +184,45 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+				fp.reportCriticalErr(err)
+				continue
+			}
+			if !hasVp {
+				// the randomness is already committed but
+				// the finality provider does not have voting power
+				// we can safely set lastProcessHeight now
+				fp.MustSetLastProcessedHeight(b.Height)
 				continue
 			}
```
SebastianElvis, line 213: Can this be moved to the front of other checks? if the fp does not have voting power at that height, then it can never be. So it seems safe to skip all the subsequent checks to go to the next block?

finality-provider/service/fastsync.go
```diff
@@ -49,19 +49,27 @@ func (fp *FinalityProviderInstance) FastSync(startHeight, endHeight uint64) (*Fa
 		// have gaps during sync
 		catchUpBlocks := make([]*types.BlockInfo, 0, len(blocks))
 		for _, b := range blocks {
-			should, err := fp.shouldSubmitFinalitySignature(b)
+			// check whether the block has been processed before
+			if fp.hasProcessed(b) {
+				continue
+			}
```
SebastianElvis, line 55: Probably a nitpick -- can `startHeight` be at least the last processed height such that we don't need to perform this check?

SebastianElvis (commented):
Some minor complimentary questions. Looks like this part is pretty tricky to handle...

finality-provider/service/fastsync.go
```diff
@@ -49,19 +49,27 @@ func (fp *FinalityProviderInstance) FastSync(startHeight, endHeight uint64) (*Fa
...
 				return nil, err
 			}
-			if !should {
-				// two cases could lead to here:
-				// 1. insufficient committed randomness
-				// 2. no voting power
-				// thus we should continue here in case the two conditions
-				// will not happen in the rest of the blocks
+			if !hasRand {
 				continue
```
vitsalis, line 70: Maybe we should not continue here but break? i.e. the fp has voting power but no randomness committed, I guess we should stop here.

finality-provider/service/fp_instance.go
```diff
@@ -184,36 +184,45 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+			hasRand, err := fp.hasRandomness(b)
 			if err != nil {
 				fp.reportCriticalErr(err)
 				continue
 			}
-
-			if !should {
-				fp.MustSetLastProcessedHeight(nextBlock.Height)
+			if !hasRand {
+				continue
```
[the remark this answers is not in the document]
vitsalis: Agreed that stalling might be a good idea. Otherwise if we don't want to stall, we could exit and revisit this block later?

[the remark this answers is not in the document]
gitferry: I'm not sure how to achieve stalling here. Did you mean we add a for loop here to periodically check if we have randomness?

finality-provider/service/fastsync.go
gitferry: yeah I think we should break here. As fast-sync will retry later

finality-provider/service/fp_instance.go
```diff
@@ -388,35 +381,50 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlock *types.BlockInfo) (*
...
+
+	return true, nil
+}
+
+func (fp *FinalityProviderInstance) hasRandomness(b *types.BlockInfo) (bool, error) {
+	lastCommittedHeight, err := fp.GetLastCommittedHeight()
+	if err != nil {
+		return false, err
+	}
+	if b.Height > lastCommittedHeight {
```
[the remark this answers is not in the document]
gitferry: Currently randomness is submitted based on the last committed height from Babylon, so no gaps regardless voting power

finality-provider/service/fastsync.go
gitferry: The poller does initiate from at least `LastProcessedHeight` but add this check just to be safe

finality-provider/service/fp_instance.go
gitferry: Good point!

finality-provider/service/fastsync.go
KonradStaniec: wouldn't `break` exit the loop and close finalty provider instance ?

finality-provider/service/fp_instance.go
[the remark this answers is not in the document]
KonradStaniec: I would say so, add some for loop equivalent (or retry.Do) which waits for either block being finalized or randomness being commited.

finality-provider/service/fastsync.go
vitsalis: `break` here would stop the `FastSync` method I think

gitferry: No. This is the fast-sync loop. Not the finality sig submission loop 😄

finality-provider/service/fp_instance.go
```diff
@@ -184,36 +184,44 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+				continue
+			}
+			// check whether the randomness has been committed
+			hasRand, err := fp.hasRandomness(b)
+			if err != nil {
+				fp.reportCriticalErr(err)
+				continue
+			}
+			if !hasRand {
 				continue
```
vitsalis, line 211: If we continue here, won't the current processed block be lost and never visited again as we have already consumed it?

finality-provider/service/fp_instance.go
```diff
@@ -184,36 +184,46 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+			if !hasVp {
+				// the finality provider does not have voting power
+				// and it will never will at this block
+				fp.MustSetLastProcessedHeight(b.Height)
 				continue
 			}
+			// check whether the randomness has been committed
+			// we should stall here until we have randomness committed at this
+			// height, otherwise, we might miss blocks
+			canProceed, err := fp.retryCheckRandomnessUntilBlockFinalized(b)
```
SebastianElvis, line 207: This functions seems to return either `true, nil` or `false, err` so probably we can allow it to just return `error`?

SebastianElvis (approved):
Lgtm, great work!

finality-provider/service/fp_instance.go
gitferry: Good point!

finality-provider/service/fp_instance.go
```diff
@@ -435,6 +440,59 @@ func (fp *FinalityProviderInstance) checkLagging(currentBlock *types.BlockInfo)
...
+				zap.Uint64("target_block_height", targetBlock.Height),
+				zap.Error(err),
+			)
+
+			failedCycles += 1
+			if failedCycles > uint32(fp.cfg.MaxSubmissionRetries) {
+				return fmt.Errorf("reached max failed cycles with err: %w", err)
+			}
+		} else if hasRand {
+			// the signature has been successfully submitted
```
KonradStaniec, line 470: `the signature has been successfully submitted` -> `the randomness has been successfully submitted`

finality-provider/service/fp_instance.go
```diff
@@ -435,6 +440,59 @@ func (fp *FinalityProviderInstance) checkLagging(currentBlock *types.BlockInfo)
...
+			if err != nil {
+				return fmt.Errorf("failed to query block finalization at height %v: %w", targetBlock.Height, err)
+			}
+			if finalized {
+				fp.logger.Debug(
+					"the block is already finalized, skip checking randomness",
+					zap.String("pk", fp.GetBtcPkHex()),
+					zap.Uint64("target_height", targetBlock.Height),
+				)
+				return nil
```
KonradStaniec, line 486: this is a bit debatable as if we are here it means that we should have voted but we did not because we failed to commit randomness. This sound as error condition, although I am not sure we can do anything about it here as this is mostly problem of randomness commitment loop. Maybe `TODO` to investigate it a bit more ?

KonradStaniec (approved):
LGTM!

finality-provider/service/fp_instance.go
gitferry: Agreed. This also applies to the randomness/finality vote submission loop. Also added an issue to track it

vitsalis: Don't think that's an error, as the block might have become finalized through other fps

finality-provider/service/fp_instance.go
```diff
@@ -435,6 +440,61 @@ func (fp *FinalityProviderInstance) checkLagging(currentBlock *types.BlockInfo)
...
+// finalized
+// error will be returned if maximum retries have been reached or the query to
+// the consumer chain fails
+func (fp *FinalityProviderInstance) retryCheckRandomnessUntilBlockFinalized(targetBlock *types.BlockInfo) error {
+	var failedCycles uint32
+
+	// we break the for loop if the block is finalized or the randomness is successfully committed
+	// error will be returned if maximum retries have been reached or the query to the consumer chain fails
+	for {
+		// error will be returned if max retries have been reached
```
vitsalis, line 454: Should we have a debug log here that we are checking randomness?

vitsalis (approved):
nice!

finality-provider/service/fp_instance.go
KonradStaniec: sure, but we did not vote on it even though we should becouse we did not have randomness commited. So we lost potential rewards. The question is why we did not have randomness committed - due to lack of funds ?due to gaslimits in blocks ?due to bug in software?. 

This seems like situation to explain/log/investigate rather than pretend everything is fine.

PR #235, opened 2024-02-05
fix(poller): Next height should not be set with a lower value

### LINKED ISSUE #234: Fast-sync inconsistent with poller

gitferry:
This PR potentially fixed #234. We need to ensure that after fast sync, the poller is still up-to-date. This PR ensures that the next height of the poller should not be set with a lower value.

### DIFF
--- a/finality-provider/service/chain_poller.go
+++ b/finality-provider/service/chain_poller.go
@@ -26,18 +26,27 @@ const (
 	maxFailedCycles = 20
 )
 
+type skipHeightRequest struct {
+	height uint64
+	resp   chan *skipHeightResponse
+}
+
+type skipHeightResponse struct {
+	err error
+}
+
 type ChainPoller struct {
 	startOnce sync.Once
 	stopOnce  sync.Once
 	wg        sync.WaitGroup
-	mu        sync.Mutex
 	quit      chan struct{}
 
-	cc            clientcontroller.ClientController
-	cfg           *cfg.ChainPollerConfig
-	blockInfoChan chan *types.BlockInfo
-	nextHeight    uint64
-	logger        *zap.Logger
+	cc             clientcontroller.ClientController
+	cfg            *cfg.ChainPollerConfig
+	blockInfoChan  chan *types.BlockInfo
+	skipHeightChan chan *skipHeightRequest
+	nextHeight     uint64
+	logger         *zap.Logger
 }
 
 func NewChainPoller(
@@ -46,11 +55,12 @@ func NewChainPoller(
 	cc clientcontroller.ClientController,
 ) *ChainPoller {
 	return &ChainPoller{
-		logger:        logger,
-		cfg:           cfg,
-		cc:            cc,
-		blockInfoChan: make(chan *types.BlockInfo, cfg.BufferSize),
-		quit:          make(chan struct{}),
+		logger:         logger,
+		cfg:            cfg,
+		cc:             cc,
+		blockInfoChan:  make(chan *types.BlockInfo, cfg.BufferSize),
+		skipHeightChan: make(chan *skipHeightRequest, 1),
+		quit:           make(chan struct{}),
 	}
 }
 
@@ -184,8 +194,8 @@ func (cp *ChainPoller) waitForActivation() {
 		if err != nil {
 			cp.logger.Debug("failed to query the consumer chain for the activated height", zap.Error(err))
 		} else {
-			if cp.GetNextHeight() < activatedHeight {
-				cp.SetNextHeight(activatedHeight)
+			if cp.nextHeight < activatedHeight {
+				cp.nextHeight = activatedHeight
 			}
 			return
 		}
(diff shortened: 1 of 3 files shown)


finality-provider/service/chain_poller.go
```diff
@@ -261,7 +261,9 @@ func (cp *ChainPoller) getNextHeight() uint64 {
 func (cp *ChainPoller) setNextHeight(height uint64) {
 	cp.mu.Lock()
 	defer cp.mu.Unlock()
-	cp.nextHeight = height
+	if cp.nextHeight < height {
```
vitsalis, line 264: Maybe the caller (e.g. `SetNextHeight`) does this check and we keep this as an internal setter?

gitferry: Good point. Done.

finality-provider/service/chain_poller.go
```diff
@@ -249,23 +249,25 @@ func (cp *ChainPoller) pollChain() {
...
 	return cp.nextHeight
 }
 
 func (cp *ChainPoller) setNextHeight(height uint64) {
-	cp.mu.Lock()
-	defer cp.mu.Unlock()
 	cp.nextHeight = height
 }
 
 func (cp *ChainPoller) SetNextHeight(height uint64) {
```
KonradStaniec, line 265: I would argue this is not correct fix due two things.

1. From api perspective, I would be suprised if I set some height and component would just ignore it

2. Looking at:
```
			cp.SetNextHeight(blockToRetrieve + 1)
			failedCycles = 0
			cp.logger.Info("the poller retrieved the block from the consumer chain",
				zap.Uint64("height", block.Height))
			// push the data to the channel
			// Note: if the consumer is too slow -- the buffer is full
			// the channel will block, and we will stop retrieving data from the node
			cp.blockInfoChan <- block
```
In scenario in which update happen mid download, old block is still pushed onto the channel. This is imo unexpected from the pov of the caller.

KonradStaniec: hmm thats one option, although sharing channels is tricky to get right and couples components.

Another option would be to have some thing like:
```
type restartRequest struct {
	height uint64
	resp   chan struct{}
}
```

and in poller:
```
restartChan chan *restartRequest
```

then you can add another case to selec in poller:
```
		select {
		case req := <-cp.restartChan:
			// drain blockInfoChan channel
			// sent next height to req.height
			req.resp <- struct{}{}

```
and have proper restart from heigh function:
```
func (cp *ChainPoller) RestartFromHeight(height uint64) {
	resp := make(chan struct{}, 1)
	cp.restartChan <- &restartRequest{height: height, resp: resp}
	<-resp
}
```

Now after restart is called, buffer will always return block higher or equal `height`. Also you can restart poller from any height.

One downside is that if we are mid `blockWithRetry` request it will wait for it to finish. This could be improved by storing `cancelFunc` in chain poller if there is request happening, but it is debatable whether this is necessary.

gitferry: Looks great. Thanks! Will implement

gitferry:
@KonradStaniec Thanks for your valuable comments. I have made changes accordingly and added unit tests. Please review it again. Thanks!

finality-provider/service/chain_poller.go
```diff
@@ -46,11 +55,12 @@ func NewChainPoller(
...
-		logger:        logger,
-		cfg:           cfg,
-		cc:            cc,
-		blockInfoChan: make(chan *types.BlockInfo, cfg.BufferSize),
-		quit:          make(chan struct{}),
+		logger:         logger,
+		cfg:            cfg,
+		cc:             cc,
+		blockInfoChan:  make(chan *types.BlockInfo, cfg.BufferSize),
+		skipHeightChan: make(chan *skipHeightRequest, 1),
```
KonradStaniec, line 62: this probably can be unbuffered, so that skip by height request do not accumulate

finality-provider/service/chain_poller.go
```diff
@@ -242,37 +252,47 @@ func (cp *ChainPoller) pollChain() {
...
+
 		case <-cp.quit:
 			return
 		}
 	}
 }
 
-func (cp *ChainPoller) GetNextHeight() uint64 {
-	return cp.getNextHeight()
+func (cp *ChainPoller) SkipToHeight(height uint64) error {
```
KonradStaniec, line 285: this needs to cover two additional cases:
1.  some one closed poller before calling `SkipToHeight`. If that happened then request will never be processed and this will hang
2. some one closed poller while waiting for response to `&skipHeightRequest`

This can be acheived by writing this function like:
```
func (cp *ChainPoller) SkipToHeight(height uint64) error {
	respChan := make(chan *skipHeightResponse, 1)

	select {
	case <-cp.quit:
		return fmt.Errorf("the chain poller is stopped")
	case cp.skipHeightChan <- &skipHeightRequest{height: height, resp: respChan}:
	}

	select {
	case <-cp.quit:
		return fmt.Errorf("the chain poller is stopped")
	case resp := <-respChan:
		return resp.err
	}
}
```
This is a bit mouthful so in oder repos we have some utils to do this like in staker - https://github.com/babylonchain/btc-staker/blob/main/utils/utils.go#L37

KonradStaniec (commented):
thanks for changes (and tests 👍 )

just two small additional comments

finality-provider/service/chain_poller.go
gitferry: Nice find! I also realized that we should return error if the poller has not been started yet. To cover these cases, I changed `startOnce/stopOnce sync.Once` to `isStarted *atomic.Bool` and added your suggestions above. Also, I have changed the tests to cover these cases. Please take a look again. Thanks!

PR #236, opened 2024-02-05
Bumps babylon to latest version

KonradStaniec:


clientcontroller/babylon.go
```diff
@@ -234,25 +225,16 @@ func (bc *BabylonController) SubmitBatchFinalitySigs(fpPk *btcec.PublicKey, bloc
...
 		finalitytypes.ErrPubRandNotFound,
 		btcstakingtypes.ErrFpAlreadySlashed,
 	}
 
-	res, err := bc.reliablySendMsgs(msgs, expectedErrs, unrecoverableErrs)
+	res, err := bc.reliablySendMsgs(msgs, emptyErrs, unrecoverableErrs)
 	if err != nil {
 		return nil, err
 	}
-	if res == nil {
```
vitsalis, line 251: hmm maybe let's leave these `if`s? It's ok that we don't have any expected errors right now, but in the future we might have and this might come to bite us.

vitsalis (approved):

clientcontroller/babylon.go
KonradStaniec: tbh in the future I would prefer to do the cleanup on rpc client side and:
1. either remove this concept of expected errors. Imo this is weird concept, as if I get error which I am expecting, should I do not the thing I am doing in the first place ?
2. or in case of expected error return some kind of wrapped error of value `ErrExpected` . 
Imo returning nil response is really most unexpected choice.

gitferry: I agree with Konrad here. The caller should expect that if the returned err is nil, the response should not be nil

gitferry (approved):
Lgtm!

PR #247, opened 2024-02-25
Add kubernetes ready dockerfile for finality provider

asutosh23:
This PR adds a Dockerfile for Finality Provider that supports all the three tools inside it such as: eotsd, fpd, fpdcli. This is specifically for users not for developers. Whereas users can use this to run a local instance of any of these 3 and use this image in kubernetes also.

- This is more user friendly rather than developer friendly.
- More secure as it uses debian all the way.
- More lightweight as the final stage has only binaries nothing else.
- The user doesn't need the to clone the whole repo with all branches locally, the dockerfile will clone selected release branch which will take less time than the other one to clone.
- It's ready to be used in k8s right away
- It has a mention in the readme, if this gets merged then I will mention in the official docs page ago, so that it will be easy to find

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,10 @@ export PATH=$HOME/go/bin:$PATH
 echo 'export PATH=$HOME/go/bin:$PATH' >> ~/.profile
 ```
 
+### Running in Docker
+
+The Dockerfile provided [here](./Dockerfile) can be used for developement of Finality Provider whereas a modifed version of the same Dockerfile, which is [this one](./docs/fp.Dockerfile)(exports with `eotsd`, `fpd`, `fpcli` preinstalled) can be used by the users to run a FP container locally or in Kubernetes. Users can paralley run 3 containers of this image  for all the three tools and run their commands inside them respectively.
+
 ## 3. Setting up a finality provider
 
 ### 3.1. Setting up a Babylon Full Node
--- a/docs/fp.Dockerfile
+++ b/docs/fp.Dockerfile
@@ -1,0 +1,63 @@
+####################
+### Build Binaries #
+####################
+
+FROM golang:1.21.1 AS build-env
+
+# Version to build. Default is the Git HEAD.
+ARG VERSION="HEAD"
+
+# Use muslc for static libs
+ARG BUILD_TAGS="muslc"
+
+RUN apt-get update && \
+  apt-get --no-install-recommends --yes install \
+  git-core \
+  wget \
+  bzip2 \
+  jq \
+  vim \
+  ca-certificates \
+  build-essential \
+  pkg-config \
+  cmake \
+  gcc
+
+
+# Cosmwasm - download correct libwasmvm version
+# fp binaries are dependent on this package to run
+RUN git clone --branch v0.1.0 https://github.com/babylonchain/finality-provider.git && \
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -80,6 +80,10 @@ export PATH=$HOME/go/bin:$PATH
 echo 'export PATH=$HOME/go/bin:$PATH' >> ~/.profile
 ```
 
+### Running in Docker
```
gitferry, line 83: Maybe we can provide some sample commands in Makefile. Currently, we have `make build-docker`. Maybe we should add `make build-docker-release` for this. And also commands for running the docker container.

README.md
```diff
@@ -80,6 +80,10 @@ export PATH=$HOME/go/bin:$PATH
...
+
+The Dockerfile provided [here](./Dockerfile) can be used for developement of Finality Provider whereas a modifed version of the same Dockerfile, which is [this one](./docs/fp.Dockerfile)(exports with `eotsd`, `fpd`, `fpcli` preinstalled) can be used by the users to run a FP container locally or in Kubernetes. Users can paralley run 3 containers of this image  for all the three tools and run their commands inside them respectively.
```
gitferry, line 85: Please ensure that each line wraps at 80 characters for better readability.

gitferry (commented):
Thanks for adding this. In general, it makes sense to me. Left some high level comments and leave this to @filippos47 for a final pass

docs/fp.Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+  ca-certificates \
+  build-essential \
+  pkg-config \
+  cmake \
+  gcc
+
+
+# Cosmwasm - download correct libwasmvm version
+# fp binaries are dependent on this package to run
+RUN git clone --branch v0.1.0 https://github.com/babylonchain/finality-provider.git && \
```
filippos47, line 29: Hardcoding a branch doesn't really make sense to me. Every time we have a new release, we'd need to update this Dockerfile.

docs/fp.Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+  BUILD_TAGS=$BUILD_TAGS \
+  LINK_STATICALLY=true && \
+  make build
+
+
+# ###########################
+# ### Finality Provider Run #
+# ###########################
+
+FROM debian:bookworm-slim AS run-env
```
filippos47, line 51: With a simple search, I see several production blockchain node Dockerfiles utilizing an Alpine base image for the final layer. Not really certain of the merits here.

filippos47 (changes requested):
The changes I see from our original Dockerfile are:
- Change base image on final container from Alpine to Debian
- Remove a few packages which we've added for debugging purposes on the final image (https://github.com/babylonchain/finality-provider/blob/14514b6a63aec3dc75a22a4cf9d86c0a02bb8469/Dockerfile#L42)
- Hardcode a Finality Provider branch inside the Dockerfile

The only part that I agree is the removal of the debugging packages. With the original Dockerfile, we've been running Finality Providers and EOTS daemons on Kubernetes without any issues.

Thus, I'd convert this PR to a one-line change in the original Dockerfile that removes the debugging packages - no need for README update either.

PR #250, opened 2024-03-04
chore: `fpcli ls` shows info of all the managed finality providers

### LINKED ISSUE #244: `fpcli ls` should return all the managed fps with different status
Current, the command `fpcli ls` only returns fps that have been registered and running within the daemon

gitferry:
This PR fixed #244. This PR allows `fpcli ls` to show info of all the managed finality providers other than only the ones that are running within the daemon.


gitferry:
It seems that the CI server is down? @filippos47

vitsalis:
@gitferry It's the same issue I encountered here https://github.com/babylonchain/babylon/pull/524. Need to update the circleci image

gitferry:
@vitsalis Thanks! CI has passed

testutil/utils.go
```diff
@@ -14,7 +15,8 @@ import (
 const TestPubRandNum = 25
 
 func EmptyDescription() []byte {
-	return []byte("empty description")
+	des := stakingtypes.NewDescription("", "", "", "", "")
```
SebastianElvis, line 18: Babylon side requires that at least the moniker should not be empty ([ref](https://github.com/babylonchain/babylon/blob/107ab1b457f7b4caef846239990fc93a23efe96a/x/btcstaking/types/msg.go#L27-L32)) so maybe let's give a random moniker here

finality-provider/service/fp_manager.go
```diff
@@ -279,6 +279,55 @@ func (fpm *FinalityProviderManager) ListFinalityProviderInstances() []*FinalityP
 	return fpisList
 }
 
+func (fpm *FinalityProviderManager) AllFinalityProviders() ([]*proto.FinalityProviderInfo, error) {
```
SebastianElvis, line 282: Since `AllFinalityProviders` and `FinalityProviderInfo` will be used for providing responses to the RPC server, could `ListFinalityProviderInstances` and `GetFinalityProviderInstance` be made private? `FinalityProviderInstance` carries some sensitive information about the finality provider so should probably not be queried by external callers.

SebastianElvis (approved):

finality-provider/service/fp_manager.go
gitferry: Good point. Currently, making these methods public is merely for e2e tests. Will need some refactoring to make it work, which is probably not the scope of this PR. Will do it in a refactoring PR

PR #255, opened 2024-03-07
feat: Fix not change the `fp`'s status to `REGISTERED` problem

choihocheol:
I encountered a problem, my `fpd` doesn’t vote.
My `fpd` did not proceed, even after waiting a day with the following log.
```bash
info    successfully connected to a remote EOTS manager    {"address": "127.0.0.1:12582"}
info    Starting FinalityProviderApp
info    RPC server listening    {"address": "[::]:12581"}
info    Finality Provider Daemon is fully active!
```

I discovered that my FP status remained ‘CREATED’ in the `fpd` database, despite already registered on Babylonchain.
I suspect the issue arose from restarting my `fpd` node before receiving a response from `finalityProviderRegisteredEventChan`.

I’ve implemented the `ForceRegisterFinalityProvider` feature in `fpd` and `fpcli`.
This feature aims to change the `fp` status from `CREATED` to `REGISTERED` in the fpd database, provided that the `fp` is already registered on the Babylon chain.
Below, you will find a CLI example demonstrating how to use this feature.
```bash
fcli register-finality-provider \
	--force true \
	--btc-pk ${the finality provider btc pk in problem}
```

I conducted a test of this feature on my `fpd` node and confirmed that my node is now voting correctly by using the query `babylond q finality votes-at-height`.


gitferry (changes requested):
Thanks for the fix! Some suggestions:
* the `force` flag and `ForceRegisterFinalityProvider` handler are probably not needed. We can check the registration status within the normal [registration handler](https://github.com/choihocheol/finality-provider/blob/69e77f0af7df6c61bce141cef5cdd03a6e91872d/finality-provider/service/app.go#L166)
* the pr title could be renamed to `fix` other than `feat`

choihocheol:
Thanks for the comments @gitferry .

- Understood. What about this implementation? Is it would be okay?
1. If `fpd` encounters the error "Babylon: finality-provider is already registered at this line” in [this code line](https://github.com/choihocheol/finality-provider/blob/69e77f0af7df6c61bce141cef5cdd03a6e91872d/finality-provider/service/app.go#L210).
	- The error indicates that the status of `fp` is still `CREATED` in `fpd DB`, even though `fp` has been registered on Babylonchain.
2. Change the status of `fp` in the `fpd DB` from `CREATED` to `REGISTERED`.

- Alright, I will update the PR title to `fix`.

gitferry:
@choihocheol I think this will mix logc and error handling. It's more intuitive to update the status proactively other than wait until an error is hit.

choihocheol:
@gitferry Okay, I see what you mean, I'll implement and PR it as you said for improving UX.

choihocheol:
I create new PR #263 regarding this problem.
So I close this PR.

PR #257, opened 2024-03-11
Hotfix: Query voting power returns `0` for future block height

gitferry:
This PR is a hotfix to resolve the issue that when rpc load balancer is used, querying voting power could return `0` which should be `>0` if the block has not been indexed. This PR fixed the issue by using `FinalityProviderCurrentPower` instead of `FinalityProviderPowerAtHeight`, which is hacky solution. A more proper solution should be from Babylon side to return an error when query voting power for non-indexed blocks.

### DIFF
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -6,9 +6,6 @@ import (
 	"time"
 
 	sdkErr "cosmossdk.io/errors"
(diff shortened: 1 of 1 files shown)


SebastianElvis:
If it's a hotfix then probably the PR can point to `main`?

gitferry:
Closing this PR as we have agreed that the hotfix is better to be on Babylon side

PR #263, opened 2024-03-15
fix: Fix not change the fp's status to REGISTERED problem

### LINKED ISSUE #253: Finality provider status `CREATED` is not changed to `REGISTERED` in some cases
We need a status sync process before starting a finality provider instance

choihocheol:
This PR fixed #253, This PR aims to change the `fp` status from `CREATED` to `REGISTERED` in the `fpd` database, provided that the `fp` is already registered on the Babylon chain.

Comparison with Previous PR #255,
- Improve UX
	- During the `fpd start`, now perform a registration check.
	- Detect are `fps` already registered on Babylon chain, and If already registered then change status in the `fpd` database.
- Detect registration via `QueryFinalityProviderSlashed`
	- The reason why I don't use `ErrFpRegistered`, the function `RegisterFinalityProvider` would do retrying 5 times and returned wrapped error.
	- So I reckon using `QueryFinalityProviderSlashed` function for detect registration is the more cleaner implementation.


gitferry (changes requested):
Thank you for your update. I don't think `QueryFinalityProviderSlashed` is a good query to use. I recommend to use `QueryFinalityProviderPowerAtHeight` but to get the block height to pass it in, you need to first use `QueryBestBlock` to get the current best height. And the logic would be:
1. if err not nil, confitue
2. if voting power > 0, set the status to `ACTIVE`
3. else if voting power == 0, set status depending on previous status
4. if previous status is `CREATED`, then set status to `REGISTERED`
5. else if previous status is `ACTIVE`, then set status to `INACTIVE`
6. else status no need to change

choihocheol:
That makes sense! Thank you for your guidance, @gitferry .
I’ve implemented the changes as you suggested and added a commit to this PR. Could you please review it?

gitferry (approved):
Great work! Lgtm!

PR #264, opened 2024-03-15
chore: Bump babylon c2c92d903a48

gitferry:
Bumps babylon version and adapt to new interfaces.

It seems that after the bump, we cannot use `error.Is` to expect error types from Babylon. This PR used `Strings.Contains` instead


gitferry:
Opps, e2e failed. Looking into it now

gitferry:
Fixed

KonradStaniec (approved):

Issue #57, opened 2023-08-03
Prometheus metrics


gitferry:

gitferry:
addressed by #265 and #269

PR #265, opened 2024-03-17
feat: Setup and add metrics

gusin13:
Fixes - https://github.com/babylonchain/finality-provider/issues/57

Full metrics list here - https://babylon-chain.atlassian.net/wiki/spaces/BABYLON/pages/203980929/Finality+Provider+Metrics

Sets up and adds following metrics

1. total_running_fps
2. fp_status
3. babylon_tip_height
4. last_polled_height
5. poller_starting_height
6. fp_seconds_since_last_vote
7. fp_seconds_since_last_randomness
8. fp_last_voted_height
9. fp_last_processed_height
10. fp_total_blocks_without_voting_power
11. fp_total_voted_blocks
12. fp_total_committed_randomness
13. fp_last_committed_randomness_height
14. fp_total_failed_votes
15. fp_total_failed_randomness

Notes:
This PR doesn't add eots related metrics yet, will do in next pr.


finality-provider/service/types.go
```diff
@@ -144,6 +144,7 @@ func (fp *FinalityProviderInstance) MustSetLastProcessedHeight(height uint64) {
 		fp.logger.Fatal("failed to set last processed height",
 			zap.String("pk", fp.GetBtcPkHex()), zap.Uint64("last_processed_height", height))
 	}
+	fp.metrics.RecordFpLastProcessedHeight(fp.GetBtcPkHex(), height)
```
gitferry, line 147: This should also be recorded within `MustUpdateStateAfterFinalitySigSubmission`

finality-provider/service/fp_instance.go
```diff
@@ -540,6 +548,9 @@ func (fp *FinalityProviderInstance) retrySubmitFinalitySignatureUntilBlockFinali
 			}
 		} else {
 			// the signature has been successfully submitted
+			fp.metrics.RecordFpVoteTime(fp.GetBtcPkHex())
```
gitferry, line 551: Maybe it's better to set it within `SubmitFinalitySignature` as it is closer to the truth

finality-provider/service/fp_instance.go
```diff
@@ -540,6 +548,9 @@ func (fp *FinalityProviderInstance) retrySubmitFinalitySignatureUntilBlockFinali
...
+			fp.metrics.RecordFpLastVotedHeight(fp.GetBtcPkHex(), targetBlock.Height)
+			fp.metrics.IncFpTotalVotedBlocks(fp.GetBtcPkHex())
```
gitferry, line 553: This should also be set within `SubmitFinalitySignature`

finality-provider/service/fp_instance.go
```diff
@@ -540,6 +548,9 @@ func (fp *FinalityProviderInstance) retrySubmitFinalitySignatureUntilBlockFinali
 			}
 		} else {
 			// the signature has been successfully submitted
+			fp.metrics.RecordFpVoteTime(fp.GetBtcPkHex())
+			fp.metrics.RecordFpLastVotedHeight(fp.GetBtcPkHex(), targetBlock.Height)
```
gitferry, line 552: The metric of last voted height should be set along with the update of `last voted height`, so a better place to set this is within `MustUpdateStateAfterFinalitySigSubmission`

finality-provider/service/fp_instance.go
```diff
@@ -604,6 +615,9 @@ func (fp *FinalityProviderInstance) retryCommitPubRandUntilBlockFinalized(target
 			}
 		} else {
 			// the public randomness has been successfully submitted
+			fp.metrics.RecordFpRandomnessTime(fp.GetBtcPkHex())
+			fp.metrics.RecordFpLastCommittedRandomnessHeight(fp.GetBtcPkHex(), targetBlock.Height)
+			fp.metrics.IncFpTotalCommittedRandomness(fp.GetBtcPkHex())
```
gitferry, line 620: Maybe it's better to set these three within `CommitPubRand` as it is closer to the truth

metrics/collectors.go
```diff
@@ -0,0 +1,266 @@
...
+
+func (m *Metrics) IncFpTotalVotedBlocks(fpBtcPkHex string) {
+	m.fpTotalVotedBlocks.WithLabelValues(fpBtcPkHex).Inc()
+}
+
+func (m *Metrics) AddToFpTotalVotedBlocks(fpBtcPkHex string, num float64) {
+	m.fpTotalVotedBlocks.WithLabelValues(fpBtcPkHex).Add(num)
+}
+
+func (m *Metrics) IncFpTotalCommittedRandomness(fpBtcPkHex string) {
```
gitferry, line 215: This should be `Add` rather than `Inc` as for each commitment we submit a list of randomness. The number of randomness can be obtained from within `CommitPubRand`

gitferry (commented):
Great work! Only comments about some metrics being misplaced

finality-provider/service/fp_manager.go
```diff
@@ -238,6 +243,7 @@ func (fpm *FinalityProviderManager) StartAll() error {
 	}
 
 	for _, fp := range storedFps {
+		fpm.metrics.RecordFpStatus(fp.GetBIP340BTCPK().MarshalHex(), fp.Status)
```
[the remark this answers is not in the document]
gusin13: sorry this was a left over from a old commit, I have removed it. I think we need not put this inside `SyncFinalityProviderStatus` or `monitorStatusUpdate` as we already have a dedicated metrics update ticker which updates status in metrics periodically [here](https://github.com/babylonchain/finality-provider/pull/265/files#diff-5a4b8fea3abbd5426c9d66555c41651ba451448bc23bd52b811125247f2d7193R256)

finality-provider/service/types.go
gusin13: `MustUpdateStateAfterFinalitySigSubmission` won't trigger if the fp doesn't have voting power [ref](https://github.com/babylonchain/finality-provider/blob/9b0f0e9142c9d5b62b33fadada0f080a34edc4ad/finality-provider/service/fp_instance.go#L201) so I'd say we need this both in 
`MustUpdateStateAfterFinalitySigSubmission` and `MustSetLastProcessedHeight`

gusin13: `MustUpdateStateAfterFinalitySigSubmission` won't trigger if the fp doesn't have voting power [ref](https://github.com/babylonchain/finality-provider/blob/9b0f0e9142c9d5b62b33fadada0f080a34edc4ad/finality-provider/service/fp_instance.go#L201) so I think we need it both in 
`MustUpdateStateAfterFinalitySigSubmission` and `MustSetLastProcessedHeight` ?

metrics/collectors.go
gusin13: ok i see, am wondering how is this info useful, as we will keep adding the len of pub randomness to this metric, isn't `fp_last_committed_randomness_height` suffice?

finality-provider/service/app.go
```diff
@@ -33,12 +35,6 @@ type FinalityProviderApp struct {
 	wg   sync.WaitGroup
 	quit chan struct{}
 
-	sentWg   sync.WaitGroup
```
gusin13, line 36: @gitferry want to check with you if you agree this is ok, I have removed the extra channels and using the same for event, registeration and metrics loop [ref](https://github.com/babylonchain/finality-provider/pull/265/files#diff-d1b387a8ca176ce2b31374566ddd374acde7ec02fb48b76075f9597bd4c6f0deR281)

finality-provider/service/fp_manager.go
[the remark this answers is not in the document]
gitferry: Yeah that could work

finality-provider/service/types.go
gitferry: yep, that was my point : )

metrics/collectors.go
gitferry: `fp_last_committed_randomness_height` cannot capture the case where the fp goes off or it runs out of funds for sending randomness. Also, the number of randomness can be changed by modifying the config file

finality-provider/service/app.go
gitferry: It's fine. Thanks for asking

gitferry (approved):
Nice work! One nit is that maybe it's better to move `metrics/config.go` to `config/metrics.go`

metrics/collectors.go
gusin13: hmm so iiuc `fp_total_committed_randomness` would be stop increasing if the fp goes off so if we plot a timeline graph we would notice sth bad has happened. 

but this indication could be retrieved from `fp_last_committed_randomness_height` or `fp_total_failed_randomness` also no? like if we plot a timeline graph `fp_last_committed_randomness_height` would start becoming a straight line instead of increasing, also `fp_total_failed_randomness` would start increasing rapidly

gitferry: For example, in a timeline graph, we see two points of `fp_last_committed_randomness_height`, `1` and `1000`, from which we don't know what happened in between. Along with `fp_total_committed_randomness`, this gap is filled to some extend. Also, the difference of `fp_total_committed_randomness` and `fp_total_voted_counts`, user can estimate how many randomness is wasted

gusin13:
cc @filippos47 - metrics port exposed, will raise a separate pr on helm/k8s repo soon.

PR #276, opened 2024-03-22
Consumer chain finality provider

maurolacy:
Based on / same as #275.


SebastianElvis (approved):
Probably it's good to add an e2e test (under `itest/`) to ensure registering a consumer chain FP works properly, but not a blocker for now :+1:

gitferry (approved):

maurolacy:
> Probably it's good to add an e2e test (under `itest/`) to ensure registering a consumer chain FP works properly, but not a blocker for now 👍

Can do in a follow-up.

PR #278, opened 2024-03-25
Consumer FP registration test

maurolacy:


finality-provider/service/fp_manager.go
```diff
@@ -293,6 +293,20 @@ func (fpm *FinalityProviderManager) ListFinalityProviderInstances() []*FinalityP
 	return fpisList
 }
 
+func (fpm *FinalityProviderManager) ListFinalityProviderInstancesForChain(chainID string) []*FinalityProviderInstance {
+	fpm.mu.Lock()
+	defer fpm.mu.Unlock()
+
+	fpisList := make([]*FinalityProviderInstance, 0, len(fpm.fpis))
+	for _, fpi := range fpm.fpis {
+		if string(fpi.GetChainID()) == chainID {
```
SebastianElvis, line 302: I saw `GetChainID()` is defined as
```go
func (fp *FinalityProviderInstance) GetChainID() []byte {
	return []byte(fp.state.getStoreFinalityProvider().ChainID)
}
```
so probably we could make it return `string`?

SebastianElvis (approved):

clientcontroller/babylon.go
```diff
@@ -586,3 +587,20 @@ func (bc *BabylonController) SubmitCovenantSigs(
 
 	return &types.TxResponse{TxHash: res.TxHash, Events: res.Events}, nil
 }
+
+// Test methods for e2e testing
+func (bc *BabylonController) RegisterConsumerChain(id, name, description string) (*types.TxResponse, error) {
```
gitferry, line 592: Would it be part of client controller interface?

itest/e2e_test.go
```diff
@@ -203,3 +203,18 @@ func TestFastSync(t *testing.T) {
...
+func TestConsumerFinalityProviderRegistration(t *testing.T) {
+	tm, _ := StartManagerWithFinalityProvider(t, 1)
+	defer tm.Stop(t)
+
+	consumerChainID := "consumer-chain-test-1"
+
+	_, err := tm.BBNClient.RegisterConsumerChain(consumerChainID, "Consumer chain 1 (test)", "Test Consumer Chain 1")
+	if err != nil {
+		t.Fatalf("failed to register consumer chain: %v", err)
+	}
```
gitferry, line 217:
```suggestion
	require.NoError(t, err)
```

itest/e2e_test.go
```diff
@@ -203,3 +203,18 @@ func TestFastSync(t *testing.T) {
...
+
+	tm.CreateFinalityProvidersForChain(t, consumerChainID, 1)
```
gitferry, line 219: I think this test also needs to check whether the chain and fps are registered on the Babylon side. For that, we might need some queries

gitferry (commented):
Nice work! One comment is that we probably need queries to ask Babylon for registration status. Local info is only half the story

itest/e2e_test.go
maurolacy: Queries are there. Will try to map them and use them, in a follow up.

PR #284, opened 2024-04-15
feat(bip-32): BIP-32 based randomness derivation

SebastianElvis:
Fixes https://github.com/babylonchain/babylon/issues/591

This PR bumps Babylon and accommodates the new BIP-32 based EOTS randomness derivation protocol. This is a major and breaking change, including the following notable changes:

- EOTS manager deterministically derives a pair of master randomness from chain ID and secret key
- Finality provider daemon maintains master public randomness in its storage
- Committing public randomness is replaced with the new randomness derivation mechanism
- fuzz/e2e tests
- docs

TODO before merging:
- [x] merge https://github.com/babylonchain/babylon/pull/616 and bump Babylon dependency


keyring/randgenerator.go
```diff
@@ -0,0 +1,24 @@
...
+	"crypto/hmac"
+	"crypto/sha256"
+
+	"github.com/babylonchain/babylon/crypto/eots"
+)
+
+// GenerateRandomness generates a random scalar with the given key and src
+// the result is deterministic with each given input
+func GenerateMasterRandPair(key []byte, chainID []byte) (*eots.MasterSecretRand, *eots.MasterPublicRand, error) {
+	// calculate the random hash of the key concatenated with chainID and height
```
RafilxTenfen, line 13: just chainID?

keyring/randgenerator.go
```diff
@@ -0,0 +1,24 @@
+package keyring
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+
+	"github.com/babylonchain/babylon/crypto/eots"
+)
+
+// GenerateRandomness generates a random scalar with the given key and src
```
RafilxTenfen, line 10:
```suggestion
// GenerateMasterRandPair ...
```

keyring/randgenerator.go
SebastianElvis: The `key` is used as the secret key of HMAC function at L14, so it is needed

eotsmanager/localmanager.go
```diff
@@ -136,32 +136,32 @@ func (lm *LocalEOTSManager) CreateKey(name, passphrase, hdPath string) ([]byte,
...
 }
 
-// TODO the current implementation is a PoC, which does not contain any anti-slasher mechanism
+// CreateMasterRandPair creates a pair of master secret/public randomness deterministically
+// from the finality provider's secret key and chain ID
+// TODO: the current implementation is a PoC, which does not contain any anti-slasher mechanism.
+// A simple anti-slasher mechanism could be that the manager remembers the tuple (fpPk, chainID, height) or
 //
-//	a simple anti-slasher mechanism could be that the manager remembers the tuple (fpPk, chainID, height) or
 //	the hash of each generated randomness and return error if the same randomness is requested tweice
```
gitferry, line 144: This TODO can be removed

eotsmanager/localmanager.go
```diff
@@ -136,32 +136,32 @@ func (lm *LocalEOTSManager) CreateKey(name, passphrase, hdPath string) ([]byte,
...
-	lm.metrics.SetEotsFpLastGeneratedRandomnessHeight(hex.EncodeToString(fpPk), float64(startHeight))
 
-	return prList, nil
+	return mpr.MarshalBase58(), nil
 }
 
 func (lm *LocalEOTSManager) SignEOTS(fpPk []byte, chainID []byte, msg []byte, height uint64, passphrase string) (*btcec.ModNScalar, error) {
-	privRand, _, err := lm.getRandomnessPair(fpPk, chainID, height, passphrase)
+	// get master secret randomness
+	msr, _, err := lm.getMasterRandPair(fpPk, chainID, passphrase)
```
gitferry, line 156: for each signing EOTS, we will need to generate the master key pair, which seems a waste, can we add a memory cache to it?

finality-provider/service/fp_instance.go
```diff
@@ -406,36 +361,40 @@ func (fp *FinalityProviderInstance) hasProcessed(b *types.BlockInfo) bool {
...
-func (fp *FinalityProviderInstance) hasVotingPower(b *types.BlockInfo) (bool, error) {
-	power, err := fp.GetVotingPowerWithRetry(b.Height)
+// canVote checks whether the finality provider can vote for the given block.
+// A finality provider can vote for a block if all of the following holds
+// - it has voting power at that height
+// - its registered epoch is BTC timestamping-finalised
+func (fp *FinalityProviderInstance) canVote(b *types.BlockInfo) (bool, error) {
+	// TODO(timestamping): ensure the finality provider's registered epoch is finalised
+	registeredEpoch := fp.state.fp.RegisteredEpoch
+	lastFinalizedEpoch, err := fp.cc.QueryLastFinalizedEpoch()
```
gitferry, line 371: I feel that we don't need to check this here (will waste a lot of rpc calls). In fact, the finality provider should not be started if the registered epoch has not been finalized. So, recommend check it within `NewFinalityProviderInstance`

gitferry (commented):
Love it! Some initial comments:

clientcontroller/interface.go
```diff
@@ -27,11 +26,8 @@ type ClientController interface {
 		pop []byte,
 		commission *math.LegacyDec,
 		description []byte,
-	) (*types.TxResponse, error)
-
-	// CommitPubRandList commits a list of EOTS public randomness the consumer chain
-	// it returns tx hash and error
```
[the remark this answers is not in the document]
SebastianElvis: These red lines are removed right?

eotsmanager/localmanager.go
SebastianElvis: 1 less TODO, yay!

eotsmanager/localmanager.go
SebastianElvis: Hmm I'm not sure about it. 

Ideally it should be managed like a secret key in a keyring. Anyone with the knowledge of a signature and the used public randomness can recover the SK in Schnorr. That means it should be kept encrypted in file system, and stay in memory as less as possible. Compared to keeping it in memory for a long time, calculating it ad hoc when accessing SK is more secure. As long as this does not raise performance concern (which I think it's the case) I would prefer this.

I will mark a TODO for managing master secret randomness in keyring.

finality-provider/service/fp_instance.go
SebastianElvis: Good point!

clientcontroller/interface.go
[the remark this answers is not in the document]
gitferry: sorry, I meant the doc string of `RegisterFinalityProvider`. I cannot leave comment there

eotsmanager/localmanager.go
gitferry: Makes sense. I don't worry about performance as EOTS manager is run in a separate daemon so the communication overhead is much larger than calculating the secrete master randomness

SebastianElvis:
Thanks @gitferry @RafilxTenfen for the comments! I have addressed all of them and please feel free to take a look again

finality-provider/service/fp_instance.go
```diff
@@ -73,7 +73,23 @@ func NewFinalityProviderInstance(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the last finalized epoch: %v", err)
+	}
+	if lastFinalizedEpoch < registeredEpoch {
+		logger.Debug(
+			"the finality provider's registered epoch is not BTC timestamped yet",
+			zap.String("finality_provider", sfp.KeyName),
+			zap.Uint64("registered_epoch", registeredEpoch),
+			zap.Uint64("last_finalized_epoch", lastFinalizedEpoch),
+		)
```
gitferry, line 90: is this log needed? As the returned error already contains all the information needed and shown to the user

gitferry (approved):
Great work! 👏

finality-provider/service/fp_manager_test.go
```diff
@@ -143,18 +143,24 @@ func newFinalityProviderManagerWithRegisteredFp(t *testing.T, r *rand.Rand, cc c
...
 		testutil.ZeroCommissionRate(),
+		mpr.MarshalBase58(),
 		keyName,
 		chainID,
 		pop.BabylonSig,
 		pop.BtcSig,
 	)
 	require.NoError(t, err)
 	err = fpStore.SetFpStatus(btcPk.MustToBTCPK(), proto.FinalityProviderStatus_REGISTERED)
+	err = fpStore.SetFpRegisteredEpoch(btcPk.MustToBTCPK(), 0)
```
RafilxTenfen, line 162:
```suggestion
        require.NoError(t, err)
	err = fpStore.SetFpRegisteredEpoch(btcPk.MustToBTCPK(), 0)
```

RafilxTenfen (approved):
LGTM

finality-provider/service/fp_instance.go
SebastianElvis: Hmm you are right, we could remove it

finality-provider/service/fp_manager_test.go
SebastianElvis: Thanks for the fix sir!

PR #286, opened 2024-04-16
chore: export FPD without bbn chain

RafilxTenfen:
- Moved fpcli files to its own package to avoid multiple files as main package
- Add new cli command `fpcli export-finality-provider` that exports a finality provider from the database 
- Add new RPC on `fpd` client
```go
// SignMessageFromChainKey signs a message from the chain keyring.
SignMessageFromChainKey(ctx context.Context, in *SignMessageFromChainKeyRequest, opts ...grpc.CallOption) (*SignMessageFromChainKeyResponse, error)
```
> Used to sign exported finality provider

PR #296, opened 2024-04-22
feat: add option to add new key from mnemonic

RafilxTenfen:
- Add option to create keys from mnemonic (chain and eots)
- Add instructions on export doc to how to recover with the mnemonic

> Note, the bad thing is the mnemonic is being sent over the GRPC client from fpd to eots
But keep in mind this would only happen if the finality provider lost the database files and the keys 

Should be reviewed after #286


eotsmanager/proto/eotsmanager.proto
```diff
@@ -46,6 +50,22 @@ message CreateKeyResponse {
...
 
+message CreateKeyFromMnemonicRequest {
+  // name is the identifier key in keyring
+  string name = 1;
+  // passphrase is used to encrypt the EOTS key
+  string passphrase = 2;
+  // hd_path is the hd path for private key derivation
+  string hd_path = 3;
+  // mnemonic is the mnemonic used to generate the key
+  string mnemonic = 4;
```
RafilxTenfen, line 61: To export finality provider for phase-1 the operator needs to store securely multiple files for months, which was not very good. To solve this situation, I tried to add an option to generate keys from mnemonic

The chain key I believe is all good, just prompt the user and generates it, but on the creation of the finality provider it sends it over the eots by the GRPC as a property in the request :/ 

So, this is not very nice 

@gitferry any idea how to solve this?

RafilxTenfen:
only create keys from mnemonic on fpd keys add, implementation done at https://github.com/babylonchain/finality-provider/pull/303

PR #303, opened 2024-04-23
feat: add recover flag to create key from mnemonic on fpd keys add

RafilxTenfen:
- Add new flag `--recover` to `fpd keys add` that prompts user input to write the mnemonic 
- Moved the config check update after the key creation to make config file creation optional for key creation


finality-provider/cmd/fpd/keys.go
```diff
@@ -74,10 +81,17 @@ func addKey(ctx *cli.Context) error {
...
+		mnemonic string
+		err      error
+	)
+
+	if ctx.Bool(recoverFlag) {
+		reader := bufio.NewReader(os.Stdin)
+		mnemonic, err = input.GetString("Enter your mnemonic", reader)
+		if err != nil {
+			return fmt.Errorf("failed to read mnemonic from stdin: %w", err)
+		}
```
gitferry, line 94: Maybe we can add a check to the mnemonic here. See https://github.com/cosmos/cosmos-sdk/blob/25aea8af8a52823553ddc34bd91f46857d7a54dd/client/keys/add.go#L291

gitferry (approved):
Nicely done! I was thinking maybe we can also add `import` and `export` commands to `keys` to complete the picture of recovering keys. A minor comment:

RafilxTenfen:
> Nicely done! I was thinking maybe we can also add `import` and `export` commands to `keys` to complete the picture of recovering keys. A minor comment:

Agree, I will add import and export in another PR

PR #318, opened 2024-05-09
feat: Split into ClientController and ConsumerController

lesterli:
## Summary

To support the EVM chain and keep Babylon as the staking hub, we need to split `ClientController` into two parts.

* `ClientController` includes the functions for the Babylon hub:
   * `RegisterFinalityProvider`
   * `QueryFinalityProviderVotingPower`
   * `QueryLastFinalizedEpoch`
   * ...
* `ConsumerController` includes the functions for the Consumer chain:
  *  `SubmitFinalitySig`
  * `QueryFinalityProviderVotingPower`
  * `QueryBlock`
  * ...

## Test

`make test `

```bash                                          
go test ./...

ok      github.com/babylonchain/finality-provider/clientcontroller      2.022s

ok      github.com/babylonchain/finality-provider/eotsmanager   7.521s
ok      github.com/babylonchain/finality-provider/eotsmanager/cmd/eotsd/daemon  6.552s
ok      github.com/babylonchain/finality-provider/eotsmanager/store     3.647s

ok      github.com/babylonchain/finality-provider/finality-provider/service     22.624s
ok      github.com/babylonchain/finality-provider/finality-provider/store       5.580s
ok      github.com/babylonchain/finality-provider/keyring       7.576s
```

### DIFF
--- a/clientcontroller/interface.go
+++ b/clientcontroller/interface.go
@@ -62,7 +62,7 @@ func NewClientController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams
 	return cc, err
 }
 
-type ConsumerClientController interface {
+type ConsumerController interface {
 
 	// SubmitFinalitySig submits the finality signature to the consumer chain
 	SubmitFinalitySig(fpPk *btcec.PublicKey, blockHeight uint64, blockHash []byte, sig *btcec.ModNScalar) (*types.TxResponse, error)
@@ -97,9 +97,9 @@ type ConsumerClientController interface {
 	Close() error
 }
 
-func NewConsumerClientController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ConsumerClientController, error) {
+func NewConsumerController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ConsumerController, error) {
 	var (
-		ccc ConsumerClientController
+		ccc ConsumerController
 		err error
 	)
 	switch chainName {
--- a/clientcontroller/interface.go
+++ b/clientcontroller/interface.go
@@ -32,6 +32,12 @@ type ClientController interface {
 
 	// Note: the following queries are only for PoC
 
+	// QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height
+	QueryFinalityProviderVotingPower(fpPk *btcec.PublicKey, blockHeight uint64) (uint64, error)
+
+	// QueryFinalityProviderSlashed queries if the finality provider is slashed
+	QueryFinalityProviderSlashed(fpPk *btcec.PublicKey) (bool, error)
+
 	// QueryLastFinalizedEpoch returns the last finalised epoch of Babylon
 	QueryLastFinalizedEpoch() (uint64, error)
 
--- a/clientcontroller/interface.go
+++ b/clientcontroller/interface.go
@@ -14,6 +14,7 @@ import (
 
 const (
 	babylonConsumerChainName = "babylon"
+	evmConsumerChainName     = "evm"
 )
 
 type ClientController interface {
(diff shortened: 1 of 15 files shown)


clientcontroller/interface.go
```diff
@@ -43,6 +38,40 @@ type ClientController interface {
 	// QueryFinalityProviderSlashed queries if the finality provider is slashed
 	QueryFinalityProviderSlashed(fpPk *btcec.PublicKey) (bool, error)
```
SebastianElvis, line 39: The `QueryFinalityProviderVotingPower` and `QueryFinalityProviderSlashed` queries also need to be performed to the consumer rather than Babylon. There might be some latency between Babylon and consumer in terms of the voting power / slashing status of a FP, and consumer will be deciding a FP's slashing status and voting power based on its local state

clientcontroller/interface.go
```diff
@@ -59,26 +88,23 @@ type ClientController interface {
...
 	QueryActivatedHeight() (uint64, error)
 
-	// QueryLastFinalizedEpoch returns the last finalised epoch of Babylon
-	QueryLastFinalizedEpoch() (uint64, error)
-
 	Close() error
 }
 
-func NewClientController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ClientController, error) {
+func NewConsumerClientController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ConsumerClientController, error) {
```
SebastianElvis, line 94: How about we create separate functions `NewBabylonClientController` and `NewEthClientController` for this rather than enumerating `chainName`? since

- `chainName` is a bit unintuitive to use
- Eth client might not use `bbnConfig` or `netParams`, but rather its own config object

finality-provider/service/app.go
```diff
@@ -36,6 +36,7 @@ type FinalityProviderApp struct {
 	quit chan struct{}
 
 	cc     clientcontroller.ClientController
+	ccc    clientcontroller.ConsumerClientController
```
SebastianElvis, line 39: This is a bit similar to `cc` so how about calling it something like `consumerCc`?

itest/test_manager.go
```diff
@@ -62,6 +62,7 @@ type TestManager struct {
 	Fpa               *service.FinalityProviderApp
 	EOTSClient        *client.EOTSManagerGRpcClient
 	BBNClient         *fpcc.BabylonController
+	BBCClient         *fpcc.BabylonConsumerController
```
SebastianElvis, line 65:
```suggestion
	ConsumerClient         *fpcc.BabylonConsumerController
```

clientcontroller/interface.go
```diff
@@ -14,6 +14,7 @@ import (
 
 const (
 	babylonConsumerChainName = "babylon"
+	evmConsumerChainName     = "evm"
```
SebastianElvis, line 17: Will all EVM chains share the same client controller or they can still differ in actual interface implementations?

SebastianElvis (commented):
Great first step on generalising client controller! Mostly small comments.

Also @gitferry could you also take a look to see if you agree with separating the interfaces for client controller? Won't take long, thanks :)

clientcontroller/interface.go
lesterli: as Babylon is the Staking hub and has the FP registration info, my understanding is that the consumer chain still needs to get the voting power / slashing info of FP from Babylon, right?

clientcontroller/interface.go
lesterli: yes, eth has its config, we changed the function to
[func NewConsumerClientController(config, logger *zap.Logger) (ConsumerClientController, error) ](https://github.com/Snapchain/finality-provider/blob/feat/SNA-419-support-evm-config/clientcontroller/interface.go#L94)

finality-provider/service/app.go
lesterli: OK

clientcontroller/interface.go
lesterli: currently, we assume all EVM chains use the same client controller.

clientcontroller/interface.go
lesterli: Based on offline discussions, it is better to move these two queries to the consumer.

gitferry (commented):
Sorry for the delay.  I generally agree that we should separate `clientcontroller` into two domains, interfaces for interacting with Babylon and interfaces for interacting with the consumer systems. But tbh I feel a little bit confused to see two controllers within the finality provider application. What about we still keep `ClientController` as a whole but it will contain `BabylonController` and `ConsumerController`?

SebastianElvis:
> But tbh I feel a little bit confused to see two controllers within the finality provider application.

This actually looks fine to me, given that a consumer FP indeed needs to talk to both Babylon (for registration) and consumer (for submitting finality signatures). Separating them gives some sense of clarity imo. Wdyt?

gitferry:
@SebastianElvis no strong opinion if you think it's cleaner to separate them, just the naming could be `BabylonController` and `ConsumerController`

SebastianElvis (approved):
Thanks, the code lgtm! CI failed though. Will ask infra side about the [Docker](https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1132/workflows/4643068f-9d8f-4937-929f-afed526772e6/jobs/2293) one, while https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1132/workflows/4643068f-9d8f-4937-929f-afed526772e6/jobs/2294 seems to be fixable in this PR

PR #319, opened 2024-05-11
feat: Add EVM config

lesterli:
## Summary

If the ChainName is `evm` in `fpd.conf`, when running `fpd start`, the new FP instance will use the `EVMConsumerController`.

## Test Plan

run `make test`

```bash
go test ./...

ok      github.com/babylonchain/finality-provider/clientcontroller      (cached)
ok      github.com/babylonchain/finality-provider/eotsmanager   7.422s
ok      github.com/babylonchain/finality-provider/eotsmanager/cmd/eotsd/daemon  10.423s
ok      github.com/babylonchain/finality-provider/eotsmanager/store     6.835s
ok      github.com/babylonchain/finality-provider/finality-provider/service     23.997s
ok      github.com/babylonchain/finality-provider/finality-provider/store       5.605s
ok      github.com/babylonchain/finality-provider/keyring       7.759s
```


clientcontroller/interface.go
```diff
@@ -97,14 +97,19 @@ type ConsumerController interface {
...
-func NewConsumerController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ConsumerController, error) {
+func NewConsumerController(config *fpcfg.Config, logger *zap.Logger) (ConsumerController, error) {
 	var (
 		ccc ConsumerController
 		err error
 	)
-	switch chainName {
-	case babylonConsumerChainName:
-		ccc, err = NewBabylonConsumerController(bbnConfig, netParams, logger)
+	switch config.ChainName {
```
SebastianElvis, line 105: Given that in practice only one of `BabylonConfig` and `EVMConfig` will be non-empty, how about we determine the chain type by checking which of them is non-empty, then instantiate the corresponding controller? This way we don't need `ChainName` in config

clientcontroller/evm_consumer.go
```diff
@@ -0,0 +1,109 @@
...
+	return &EVMConsumerController{
+		ec,
+		evmCfg,
+		btcParams,
+		logger,
+	}, nil
+}
+
+// SubmitFinalitySig submits the finality signature
+func (ec *EVMConsumerController) SubmitFinalitySig(fpPk *btcec.PublicKey, blockHeight uint64, blockHash []byte, sig *btcec.ModNScalar) (*types.TxResponse, error) {
```
SebastianElvis, line 44: We could use 

```suggestion
func (ec *EVMConsumerController) SubmitFinalitySig(fpPk *btcec.PublicKey, blockHeight uint64, blockHash []byte, sig *bbn.SchnorrEOTSSig) (*types.TxResponse, error) {
```

where `*bbn.SchnorrEOTSSig` is defined [here](https://github.com/babylonchain/babylon/blob/a3169a3339846d7cd09fdf7b7ffe977901dbf089/types/btc_schnorr_eots.go#L11-L12)

SebastianElvis (commented):
Nice!

clientcontroller/evm_consumer.go
gitferry: The interface requires `*btcec.ModNScalar`, doesn't it?

clientcontroller/interface.go
```diff
@@ -97,14 +97,19 @@ type ConsumerController interface {
...
+	case BabylonConsumerChainName:
+		ccc, err = NewBabylonConsumerController(config.BabylonConfig, &config.BTCNetParams, logger)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create Babylon rpc client: %w", err)
+		}
+	case EVMConsumerChainName:
+		ccc, err = NewEVMConsumerController(config.EVMConfig, &config.BTCNetParams, logger)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create Babylon rpc client: %w", err)
```
gitferry, line 114:
```suggestion
			return nil, fmt.Errorf("failed to create EVM rpc client: %w", err)
```

clientcontroller/evm_consumer.go
SebastianElvis: Hmm looks like finality provider uses `*btcec.ModNScalar` for finality signatures everywhere. Given that it's not good to enforce every consumer to depend on Babylon's struct definition let's use `*btcec.ModNScalar` then :+1:

clientcontroller/interface.go
lesterli: For the Ethereum rollup chain, both `BabylonConfig` for initializing the `BabylonController` and `EVMConfig` for initializing the `EVMConsumerController` are required. So I think we can still keep the `ChainName` in the config file.

SebastianElvis: Hmm right, okay then

finality-provider/config/evm.go
```diff
@@ -0,0 +1,15 @@
+package config
+
+const (
+	defaultEVMRPCAddr = "http://127.0.0.1:8545"
+)
+
+type EVMConfig struct {
```
gitferry, line 7: Maybe we can have a `Validate()` function to make sure the configs are in the correct format?

clientcontroller/evm_consumer.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func NewEVMConsumerController(
+	evmCfg *fpcfg.EVMConfig,
+	btcParams *chaincfg.Params,
+	logger *zap.Logger,
+) (*EVMConsumerController, error) {
+	ec, err := rpc.Dial(evmCfg.RPCAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create EVM RPC client: %w", err)
```
gitferry, line 33:
```suggestion
		return nil, fmt.Errorf("failed to connect to the EVM RPC server %s: %w", evmCfg.RPCAddr, err)
```

clientcontroller/evm_consumer.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"go.uber.org/zap"
+)
+
+var _ ConsumerController = &EVMConsumerController{}
+
+type EVMConsumerController struct {
+	evmClient *rpc.Client
+	cfg       *fpcfg.EVMConfig
+	btcParams *chaincfg.Params
```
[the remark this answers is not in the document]
lesterli: Got it.

finality-provider/config/evm.go
lesterli: Yes, `Validate()` is needed.

clientcontroller/interface.go
```diff
@@ -97,17 +97,22 @@ type ConsumerController interface {
...
-func NewConsumerController(chainName string, bbnConfig *fpcfg.BBNConfig, netParams *chaincfg.Params, logger *zap.Logger) (ConsumerController, error) {
+func NewConsumerController(config *fpcfg.Config, logger *zap.Logger) (ConsumerController, error) {
 	var (
 		ccc ConsumerController
 		err error
 	)
-	switch chainName {
-	case babylonConsumerChainName:
-		ccc, err = NewBabylonConsumerController(bbnConfig, netParams, logger)
+	switch config.ChainName {
```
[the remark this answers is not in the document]
lesterli: So for `ClientController`, there will only be the `BabylonClientController`, right?
If yes, we need to refactor the function and remove the `chainName string` for it.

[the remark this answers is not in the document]
gitferry: Ah right. We only have one option, i.e., so we can get rid of `switch`

clientcontroller/interface.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 const (
-	babylonConsumerChainName = "babylon"
-	evmConsumerChainName     = "evm"
+	BabylonConsumerChainName = "babylon"
+	EVMConsumerChainName     = "evm"
```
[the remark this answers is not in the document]
lesterli: The value `babylon` or `evm` is shorter but clear enough as we rename `ChainName` to `ConsumerChainName` in the config file.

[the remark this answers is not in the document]
gitferry: Sounds good

clientcontroller/interface.go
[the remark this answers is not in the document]
lesterli: > Maybe we can rename this config as `ConsumerChainName`

This is clear with the comment in `fpd.conf`, the new update would bring breaking changes to the running FP.
```bash
; the name of the consumer chain
ChainName = babylon
```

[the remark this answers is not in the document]
gitferry: I see. All good then

gitferry (approved):
Thanks for the change! LGTM!

PR #321, opened 2024-05-13
Release v0.1.1

RafilxTenfen:
Created release branch from dev, update the changelog with `./scripts/update_changelog.sh --token xxx --release-branch dev --since-tag v0.1.0 --future-release v0.1.1`


gitferry (commented):
Should it be `v0.2.0`? cc @vitsalis 

Also enabled `merge commit`. Please use it when merging.

maurolacy (commented):
This looks structurally good to me.

A lot of changes to review them all.

This will be merged back to `dev` from `main`, right?
Perhaps a better option is, to create a release PR against `dev` (with the CHANGELOG and version changes, etc.), merge that to `dev`, and then blindly and directly merge `dev` to `main` with a merge commit.

Just saying, since reviewing all this could be good for safety but it's a pain in the ass. And generally unfeasible, as most if not all of us lack the entire context / scope of the changes.

vitsalis (changes requested):
Two things:

- This increases the patch version which we reserve only for hotfixes. If we want to create a new release, we should create v0.2.0.
- A new release PR should only contain commits that are already in the `dev` branch. Here it looks like the changelog update is not in the `dev` branch.

maurolacy:
> A new release PR should only contain commits that are already in the dev branch. Here it looks like the changelog update is not in the dev branch.

Agreed. That will spare the merge back from `main` btw. Changing the base of this PR to `dev` will do AFAIK.

RafilxTenfen:
closed to update changelog on dev first https://github.com/babylonchain/finality-provider/pull/323

maurolacy:
Hmm, have you tried changing the base here?

RafilxTenfen:
> Hmm, have you tried changing the base here? That can be done under `Edit`.

I also modified the command to run specifying `future-release v0.2.0`

maurolacy:
You can do that with `git reset` and then force push. Anyway, the end result is the same.

PR #328, opened 2024-05-16
InfStones Finality Provider information

adam-infstones:


gitferry:
The registration should go to https://github.com/babylonchain/networks. Sorry for the confusion.

PR #335, opened 2024-05-18
split ClientController and ConsumerController (Part 2)

bap2pecs:
## Summary

> this is to follow up with https://github.com/babylonchain/finality-provider/pull/318/ to add a few missing things:

- modify and run `make mock-gen` to generate the updated mocks properly
- fix tests accordingly
    - I also modified the `make test` command so it always enforces running `make mock-gen` first
- `QueryFinalityProviderVotingPower` queries the voting power of the FP at a given consumer chain height. So this interface function shouldn't existing in both the Babylon and Consumer chain controllers. Although the source of truth of VP is on Babylon chain, there is no information about the consumer chain height. So we can only rely on the oracle and keep the interface in the consumer chain controller.
- `QueryFinalityProviderSlashed` also doesn't need to exist in both controllers. Since this function only queries the latest status of the FP, for simplicity, we can just rely on the Babylon controller's implementation since all FPs are registered and managed on the Babylon chain.
---

I also put some pseudo code in the EVM consumer controller's:

<details>
<summary>QueryFinalityProviderVotingPower</summary>

This function is only used in two places:

1. fp_instance.go when `fp.hasVotingPower(b)` is called to check whether the finality provider has voting power at a certain consumer chain's block height
2. app.go to get the VP at the latest consumer chain's block height to set the FP's status.

Since the consumer chain is a L2 rollup but the oracle contract that contains the voting power is on L1, it's impossible to get the accurate VP at the exact L2 block. So we can only aim to get the closest data.

For 2, we can just query the VP from the L1 oracle contract at the latest block in L1.

For 1, it's trickier because the sequencer only submits to L1's `L2OutputOracle` contract every X L2 blocks (i.e. 1800 for Optimism mainnet). A good estimation is to find the first submission where the submitted L2 block number is not smaller than the queried L2 height. Then get the L1 block number from the emitted event at this submission and query the voting power at this L1 block number.

We have to use "ceiling" instead of "flooring" because there can be a situation that the voting power is not added at the queried L2 block's timestamp. The only down side of using "ceiling" is the FP might miss some incentives by not signing a few blocks until the "ceiled" L2 block. But that's ok and won't sacrifice any security.

</details>

<details>
<summary>QueryActivatedHeight</summary>

In `btcstaking/types/query.pb.go`, it has 

> ActivatedHeight queries the height when BTC staking protocol is activated, i.e., the first height when there exists 1 finality provider with voting power

In our case, the oracle contract for the voting power in on L1. So similarly, we need to use it in combination with the `L2OutputOracle` contract to get the data.

Since we will emit the L1’s block number in the oracle contract when a FP’s voting power is set, we just look for the first event that contains the FP and get the L1 block number `N` from the event.

Then we just query the `L2OutputOracle`'s `OutputProposed(_outputRoot, nextOutputIndex(), _l2BlockNumber, block.timestamp, block.number)` event

to find the first event whose `block.number >= N`.

Then get the L2 block number `M` in the event and return `M`.

If cannot find such event, read from the contract's `[nextBlockNumber](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/L2OutputOracle.sol#L306C14-L306C31)()` and return the result.

Because this function `QueryActivatedHeight` is used by the chain poller module in the FP to decide which L2 block to start processing, it's ok to have some delay after the voting power is set.

</details>

<details>
<summary>QueryBestBlock</summary>
After analyzing all the call sites, I found that the purpose of the function is just to get the latest Consumer chain block number. So we just need to read it from the L2 RPC.
</details>

## Test Plan

```
make mock-gen
make test
```
for e2e-test, wait for ci/circleci


clientcontroller/interface.go
```diff
@@ -65,9 +62,6 @@ type ConsumerController interface {
 	// QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height
 	QueryFinalityProviderVotingPower(fpPk *btcec.PublicKey, blockHeight uint64) (uint64, error)
```
SebastianElvis, line 64: Certain consumers might want to minimise the communication between Babylon and FP. Probably we could still keep this interface for such consumers? It's just that the Optmism/Ethereum case does not use it.

Another reason is that slashing happens first on consumer, and then is propagated to Babylon. In some sense consumers have more updated information about slashed fps.

clientcontroller/interface.go
SebastianElvis, line 1: Probably not for this PR, I start to feel `ClientController` is not a good name anymore -- this client has to always talk to Babylon but the name is pretty generic. cc @gitferry

SebastianElvis (approved):
Lgtm!

clientcontroller/interface.go
bap2pecs: @SebastianElvis the problem is `QueryFinalityProviderVotingPower` take the consumer height as input. If the consumer wants to minimise the communication, they can just implement the logic in the consumer controller.

The point of this change is to just remove duplicate code. Right now, babylon.go and babylon_consumer.go has exactly same implementation.

It's also semantical more accurate b/c voting is for consumer chain blocks.

> Another reason is that slashing happens first on consumer, and then is propagated to Babylon. In some sense consumers have more updated information about slashed fps.

that sounds a reason we should only keep `QueryFinalityProviderVotingPower` in the consumer controller not the client controller?

clientcontroller/interface.go
bap2pecs: yes we were discussing in #318 to rename it to `BabylonController`. we have a backlog linear task for it

clientcontroller/interface.go
SebastianElvis: Sorry the comment was on whether we should remove `QueryFinalityProviderSlashed` in the consumer controller rather than for `QueryFinalityProviderVotingPower`. Somehow Github does not render the removed code in the comment :sweat:. 

I was saying that we might want to keep `QueryFinalityProviderSlashed` in teh consumer controller since some consumer wants fp to get this information from consumer side directly. Wdyt?

SebastianElvis: As per offline discussion, we agreed to remove `QueryFinalityProviderSlashed` in consumer controller now, but added a TODO in case if we want to reintroduce it in the future.

bap2pecs: discussed w @SebastianElvis offline:

> I'd vote to only add it if later we hv a consumer that needs it. for now keeping it while not used sounds wrong to me
> This sounds good to me as well, but probably let's add a TODO here so that other people working on this code base is aware of that?

conclusion:
- remove it
- add a comment so that other people understand if want to get the information from consumer side directly, they should add the interface (i will also link this discussion thread in the comment)

PR #336, opened 2024-05-19
Refactor QueryBestBlock() to only return the uint64 block height

bap2pecs:
## Summary
After analyzing all the call sites of `QueryBestBlock()`, I realized that none of the call sites needs to know the latest block's hash or whether it's finalized or not.

Plus, this function returns the latest consumer chain block, which is ofc not finalized. So the `finalized` field should always be `false`. So this function can cause confusion.

And to better reflect what this function does, I also rename to `QueryLatestBlockHeight`

## Test Plan
```
make mock-gen
make test
```


SebastianElvis (commented):
I think this is a reasonable change for Babylon/Optimism/Ethereum and other blockchains in general. However, Babylon also aims to serve some other consumers, which might not have the concept of height/block/etc.. For example, some DAG protocols might use some combinations of block hash and height as the identifier, and centralised services might use timestamps.

Thus I suggest we keep it as is for now, and generalise later once we have better ideas on the demand from the consumer side. Wdyt?

bap2pecs:
> some DAG protocols might use some combinations of block hash and height as the identifier

got it. in this case we can keep it as it is for now. but it's worth noting that before we add the support for DAG, only the `Height` field will be useful in the returned object, which is the reason I want to do this refactor --- only returns things you will actual use

bap2pecs:
my discussion w @SebastianElvis offline:

> if we r gonna add the DAG support in the near term, we can keep it as it is. but if it's not going to happen in 6 months, i'd vote to tighten it up. because for people who come later to add DAG support, it's also not guaranteed they understand this field is reserved for them. they might just add something else for the same purpose, which is even worse

> I don't think there will be many lines of development for supporting various consumers. No problem with me for tightening the interface. Let's wait for Gai's comment on Github (I already marked him as reviewer for both PRs)

will wait for the code owner @gitferry to take a look

gitferry (approved):
I like this refactoring in the sense that in `QueryBestBlock` only the height is used. For cases where the block content is needed, one can use `QueryLatestBlockHeight` + `QueryBlock`.

PR #337, opened 2024-05-19
Add QueryIsBlockFinalized() and remove BlockInfo.Finalized

bap2pecs:
## Summary

After checking how `QueryBlock()` and `BlockInfo` is used, I found that the only place that uses the `finalized` field is in fp_instance.go

so to make the code cleaner so we can later easily add the OP stack L2 consumer chain support, I am making these changes:

- add `QueryIsBlockFinalized()`
- remove the `finalized` field in `BlockInfo`

## Test Plan

```
make mock-gen
make test
```


clientcontroller/interface.go
```diff
@@ -65,8 +65,8 @@ type ConsumerController interface {
 	// QueryLatestFinalizedBlocks returns the latest finalized blocks
 	QueryLatestFinalizedBlocks(count uint64) ([]*types.BlockInfo, error)
 
-	// QueryBlock queries the block at the given height
-	QueryBlock(height uint64) (*types.BlockInfo, error)
+	// QueryIsBlockFinalized queries if the block at the given height is finalized
+	QueryIsBlockFinalized(height uint64) (bool, error)
```
SebastianElvis, line 69: This function is great as an utility function, but I'm not sure if it's expressive/generic enough as an interface. Also it's less straightfoward to use `QueryBlocks` with `limit=1` to query a certain block. Thus I suggest we kees the original interface. Wdyt?

bap2pecs: > Also it's less straightfoward to use QueryBlocks with limit=1 to query a certain block

agree. let's keep both `QueryBlocks` and `QueryBlock` then

---

but i think `QueryIsBlockFinalized` is also necessary as different consumer chain will have different implementations to determine whether a block is finalized.

SebastianElvis: Agreed. How about we add `QueryIsBlockFinalized` function into the interface, without changing existing ones then?

bap2pecs: discussed w @SebastianElvis offline. confirmed we should also remove the `finalized` field in `BlockInfo`:

> I'd vote to only add it if later we hv a consumer that needs it. for now keeping it while not used sounds wrong to me. 
> it actually caused some confusion already when we were designing the implementations. because we need to fill that field for almost all interface functions. but in fact it's only used in one place.
> my point is each API should be clear about what it needs and what it serves. serving unused data increases maintenance difficulty and decreases codebase readability (even it's just two consumer types i still need to follow every path to understand that it's actually unused field)
> if we continue to add more consumers types, new team members will be v hard to understand it's really just a field that we can fill in any value. so they will spend time to think hard how to fill that field correctly which is not needed in the first place

gitferry: I think it is good to have `QueryIsBlockFinalized` and remove `Finalized` from `BlockInfo`. This gives flexibility to the consumer chains. Also, let's keep `QueryBlock`

finality-provider/service/fp_instance.go
```diff
@@ -466,12 +466,12 @@ func (fp *FinalityProviderInstance) retrySubmitFinalitySignatureUntilBlockFinali
 }
 
 func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool, error) {
-	b, err := fp.consumerCon.QueryBlock(height)
+	finalized, err := fp.consumerCon.QueryIsBlockFinalized(height)
```
gitferry, line 469: we can return `fp.consumerCon.QueryIsBlockFinalized(height)` or we can get rid of `checkBlockFinalization`

bap2pecs:
thanks @SebastianElvis and @gitferry for the review! to summarize, we are aligned on:

- keep `QueryIsBlockFinalized`
- keep `QueryBlock`
- remove `Finalized` from `BlockInfo`

finality-provider/service/fp_instance.go
bap2pecs: good point. `checkBlockFinalization` is no longer needed. let's remove it to keep the codebase simpler

bap2pecs:
all comments resolved. and tests passed

SebastianElvis (approved):
Thanks!

PR #339, opened 2024-05-21
test base

gusin13:
This is a new base branch branched off from dev. I cherry-picked the consumer chain related commits on top. The history is much cleaner now. Once you agree, I'll need to discard the old base and then rename my branch as the new base.


gusin13:
pointed base head to latest commit of this branch, looks good now. Closing this.

PR #341, opened 2024-05-22
`QueryLatestFinalizedBlocks(count uint64)` -> `QueryLatestFinalizedBlock()`

bap2pecs:
## Summary

this changes `QueryLatestFinalizedBlocks(count uint64)` to `QueryLatestFinalizedBlock()`

after analyzing, it's only used in `latestFinalizedBlocksWithRetry()` and it's just checking 1 block. i.e. `QueryLatestFinalizedBlocks(1)`

I also don't see we need to get N latest finalized block in the near future.

so to make our life easier when we implement the interface function for evm_consumer, i am first making this change

## Test Plan

```
make mock-gen
make test
```

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -325,11 +325,11 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlockHeight uint64) (*Fast
 	}
 
 	// get the last finalized height
-	lastFinalizedBlocks, err := fp.consumerCon.QueryLatestFinalizedBlocks(1)
+	lastFinalizedBlock, err := fp.consumerCon.QueryLatestFinalizedBlock()
 	if err != nil {
 		return nil, err
 	}
-	if lastFinalizedBlocks == nil {
+	if lastFinalizedBlock == nil {
 		fp.logger.Debug(
 			"no finalized blocks yet, no need to catch up",
 			zap.String("pk", fp.GetBtcPkHex()),
@@ -338,7 +338,7 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlockHeight uint64) (*Fast
 		return nil, nil
 	}
 
-	lastFinalizedHeight := lastFinalizedBlocks[0].Height
+	lastFinalizedHeight := lastFinalizedBlock.Height
 	lastProcessedHeight := fp.GetLastProcessedHeight()
 
 	// get the startHeight from the maximum of the lastVotedHeight and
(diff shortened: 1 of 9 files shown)


bap2pecs:
will fix test later

finality-provider/service/fp_instance.go
```diff
@@ -584,13 +584,13 @@ func (fp *FinalityProviderInstance) getPollerStartingHeight() (uint64, error) {
...
 	initialBlockToGet := fp.GetLastProcessedHeight()
-	latestFinalisedBlock, err := fp.latestFinalizedBlocksWithRetry(1)
+	latestFinalisedBlock, err := fp.latestFinalizedBlockWithRetry()
 	if err != nil {
 		return 0, err
 	}
-	if len(latestFinalisedBlock) != 0 {
-		if latestFinalisedBlock[0].Height > initialBlockToGet {
-			initialBlockToGet = latestFinalisedBlock[0].Height
+	if latestFinalisedBlock != nil {
```
gitferry, line 591: Do we need to have a contract that `nil` will be returned if the latest finalized block does not exist?

finality-provider/service/fp_instance.go
```diff
@@ -325,11 +325,11 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlockHeight uint64) (*Fast
 	}
 
 	// get the last finalized height
-	lastFinalizedBlocks, err := fp.consumerCon.QueryLatestFinalizedBlocks(1)
+	lastFinalizedBlock, err := fp.consumerCon.QueryLatestFinalizedBlock()
```
gitferry, line 328: we can probably use `latestFinalizedBlockWithRetry()` here

itest/test_manager.go
```diff
@@ -340,23 +340,23 @@ func (tm *TestManager) WaitForFpVoteCast(t *testing.T, fpIns *service.FinalityPr
 	return lastVotedHeight
 }
 
-func (tm *TestManager) WaitForNFinalizedBlocks(t *testing.T, n int) []*types.BlockInfo {
+func (tm *TestManager) WaitForNFinalizedBlocks(t *testing.T, n int) *types.BlockInfo {
```
gitferry, line 343:
```suggestion
func (tm *TestManager) WaitForNFinalizedBlock(t *testing.T) *types.BlockInfo {
```

gitferry (commented):
Great work! CI failure is very easy to fix. You may want to remove
```
//go:build e2e
// +build e2e
```
on the top of `e2e.go` to see where it goes wrong but pls don't forget to add it back

itest/test_manager.go
SebastianElvis: `N` in the function name is also not needed?

finality-provider/service/fp_instance.go
bap2pecs: @gitferry what do you mean by "contract"? do you mean `latestFinalizedBlockWithRetry()`? what should be returned if not `nil` for that case?

gitferry: I mean currently we do `nil` check everywhere after calling `latestFinalizedBlockWithRetry()`. It would be good to add a note to `latestFinalizedBlockWithRetry()` and `QueryLatestFinalizedBlock` saying that `nil will be returned if the finalized block does not exist`.

itest/test_manager.go
bap2pecs: this is still needed e2e. but i am fixing the implementations

also changing the type from `int` to `uint`

bap2pecs:
close b/c this branch is named with "base" so i cannot update it

made a new branch and see the updates in #347

Issue #344, opened 2024-05-22
make sure `mock-gen` always runs when there is an interface update


bap2pecs:
#318 has some issue that i later fixed in #335 

but the CI/CD pipeline failed to detect it b/c it didn't run `make mock-gen`

bap2pecs:
<img width="860" alt="image" src="https://github.com/babylonchain/finality-provider/assets/111917526/4d016820-29e5-496e-8864-b75ac6b21d64">

@gitferry @SebastianElvis found another miss. probably need to prioritize this.

two solutions:
1) add a Git hook to always run `make mock-gen` before a push to remote
2) add a test to run `make mock-gen` and throw if the old and new file doesn't match

SebastianElvis:
Agreed to 1) run `make mock-gen` before `make test` and 2) check mock objects in CI pipeline. Probably let's do 1) first (this is immediately actionable) then do 2) at a later point (this needs some scripting)?

bap2pecs:
actually i was talking to @gitferry on my PR and we agreed that 1) is actually the thing we should avoid. CI should not run mockgen itself but alert if the author forgot to run it

bap2pecs:
git hook IMO is the easiest sol

PR #348, opened 2024-05-22
feat: support op-stack based L2 consumer chain

bap2pecs:
## Summary

this is just to consolidate the two work streams. after this, we will all work on top of `base/consumer-chain-support` branch

## Test Plan

```
make test
make test-e2e
```


Makefile
```diff
@@ -55,6 +55,7 @@ build-docker:
 .PHONY: build build-docker
 
 test:
+	make mock-gen
```
gitferry, line 58: Maybe I was missing something. Why do we want it here?

clientcontroller/interface.go
```diff
@@ -5,15 +5,15 @@ import (
...
 	"github.com/babylonchain/finality-provider/types"
 )
 
 const (
-	babylonConsumerChainName = "babylon"
+	BabylonConsumerChainName = "babylon"
+	EVMConsumerChainName     = "evm"
 )
 
 type ClientController interface {
```
gitferry, line 19: Since this Controller only works for Babylon, why not call it `BabylonController`?

gitferry (commented):
One major comment is that maybe we can rename `ClientController` to `BabylonController` as it is only for Babylon. I appologise for not thinking it deep enough in the first look

Makefile
bap2pecs: see https://github.com/babylonchain/finality-provider/pull/335 and https://github.com/babylonchain/finality-provider/issues/344

bap2pecs:
thanks for the review @gitferry! the naming of ClientController can be in another PR as it's a major refactor :)

Makefile
gitferry: We used to manually run `make mock-gen` and the generated code should be part of PR. If this is automatic in CI, we might miss updating mocks

gitferry (approved):

Makefile
bap2pecs: i see. https://github.com/babylonchain/finality-provider/pull/335 actually reflected an issue due to CI doesn't automatically run mock-gen

but agree that putting it in mock-gen seems wrong. i can remove it

SebastianElvis (approved):
1 less base branch to maintain :100:

Makefile
bap2pecs: updated

PR #352, opened 2024-05-23
chore: add check-mock-gen to the CI

bap2pecs:
## Summary

#344

## Test Plan

modified the interface temporarily, wait for the CI to finish in this PR

image

then run `make mock-gen` locally and push again

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,12 @@ jobs:
       - checkout
       - go/install:
           version: "1.21.4"
+      - go/load-cache:
+          key: go-mod-v6-{{ checksum "go.sum" }}
       - go/mod-download
+      - go/save-cache:
+          key: go-mod-v6-{{ checksum "go.sum" }}
+          path: "/home/circleci/.go_workspace/pkg/mod"
       - run:
           name: Run make mock-gen
           command: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -5,6 +5,27 @@ orbs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: |
+            go mod download
+      - run:
+          name: Run make mock-gen
+          command: |
+            make mock-gen
```
SebastianElvis, line 20: Since we are doing this, are we interested in executing `make proto-gen` to see whether protobuf fiiles are updated or not? not for this PR though

clientcontroller/interface.go
```diff
@@ -65,6 +65,8 @@ type ClientController interface {
 	Close() error
 }
 
+type ConsumerController interface{}
```
SebastianElvis, line 68: What's this interface?

bap2pecs: this is just to trigger the CI to fail. will remove before merge

SebastianElvis (approved):

.circleci/config.yml
bap2pecs: https://github.com/babylonchain/finality-provider/issues/353

PR #354, opened 2024-05-24
feat(finality): merkle-tree-based public randomness commitment

### LINKED ISSUE #350: finality: merkle tree based randomness commitment

SebastianElvis:
Closes #350 

This PR implements Merkle tree-based public randomness commitment. This includes

- reverting the BIP-32 thing that is vulnerable
- introducing storage of Merkle proofs 
- revising semantics of submitting public randomness and finality signatures

Future works that will be done in subsequent PRs:

- **DB schema:** instead of storing all proofs (which is O(n log n)), we need to find an efficient way to store Merkle tree and keep it in memory when it's used.
- **Optimisation of proof generation/veirfictaion:** we are considering to use pollarding of Merkle tree, i.e., babylon side remembers the top X levels in the Merkle tree and the Merkle proof will become hashses in the levels lower than X. see https://github.com/wealdtech/go-merkletree/blob/master/pollard.go
- **Better vector commitment schemes:** Merkle tree has other variants that have better performance in scenarios with millions of leaves. Ethereum community has stuff like Verkle tree and SSZ


finality-provider/service/eots_manager_adapter.go
```diff
@@ -0,0 +1,74 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return pubRandList, nil
+}
+
+func (fp *FinalityProviderInstance) signPubRandCommit(startHeight uint64, numPubRand uint64, commitment []byte) (*schnorr.Signature, error) {
+	msg := &ftypes.MsgCommitPubRandList{
```
gitferry, line 29: Do we need to put the finality provider pk as part of the message?

finality-provider/service/eots_manager_adapter.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, fmt.Errorf("failed to sign the commit public randomness message: %w", err)
+	}
+
+	// sign the message hash using the finality-provider's BTC private key
+	return fp.em.SignSchnorrSig(fp.btcPk.MustMarshal(), hash, fp.passphrase)
+}
+
+func (fp *FinalityProviderInstance) signFinalitySig(b *types.BlockInfo) (*bbntypes.SchnorrEOTSSig, error) {
+	// build proper finality signature request
+	msg := &ftypes.MsgAddFinalitySig{
```
gitferry, line 45: Shall we add some TODOs for the missing parts in the message?

finality-provider/cmd/fpcli/daemon/export.go
```diff
@@ -124,6 +116,9 @@ func exportFp(ctx *cli.Context) error {
 			Details:         desc.Details,
 		},
 		Commission: &comm,
+		BtcPk:      fpPk,
+		BabylonPk:  cosmosPubKey,
+		Pop:        nil, // TODO: fill PoP?
```
RafilxTenfen, line 121: :+1:

finality-provider/service/app.go
```diff
@@ -345,7 +354,54 @@ func (app *FinalityProviderApp) CreateFinalityProvider(
...
+		}
+		chainSk = &secp256k1.PrivKey{Key: keyInfo.PrivateKey.Serialize()}
+	}
+	chainPk := &secp256k1.PubKey{Key: chainSk.PubKey().Bytes()}
+
+	// 2. create EOTS key
+	fpPkBytes, err := app.eotsManager.CreateKey(req.keyName, req.passPhrase, req.hdPath)
+	if err != nil {
+		return nil, err
+	}
```
RafilxTenfen, line 377: Should we try to get the key before the call create? like if the FP had already registered that key in eots and is trying to use it 

This could happen in the context of the transition to start the chain with finality providers
This could also be done in another PR...

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+
+	// sign the commitment
+	schnorrSig, err := fp.signPubRandCommit(startHeight, numPubRand, commitment)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign the Schnorr signature: %w", err)
+	}
+
+	res, err := fp.cc.CommitPubRandList(fp.GetBtcPk(), startHeight, numPubRand, commitment, schnorrSig)
+	if err != nil {
+		// TODO Add retry. check issue: https://github.com/babylonchain/finality-provider/issues/34
```
RafilxTenfen, line 696: Should we reopen? #34

finality-provider/store/pub_rand.go
```diff
@@ -0,0 +1,136 @@
...
+			return ErrCorruptedPubRandProofDb
+		}
+
+		for i := range pubRandBytesList {
+			// check public randomness first to avoid duplicates
+			if bucket.Get(pubRandBytesList[i]) != nil {
+				return ErrDuplicatePubRand
+			}
+			// set to DB
+			if err := bucket.Put(pubRandBytesList[i], proofBytesList[i]); err != nil {
```
gitferry, line 69: Would it be good to use `height` as the key? And we can have `height -> {pubRand, proof}`. It could be helpful for range query and get the last committed height in case we need them in the future

finality-provider/service/pub_rand_store_adapter.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"sync"
+
+	"github.com/babylonchain/finality-provider/finality-provider/store"
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/cometbft/cometbft/crypto/merkle"
+)
+
+type pubRandState struct {
+	sync.Mutex
```
gitferry, line 12: Isn't the store thread-safe?

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+// retryCommitPubRandUntilBlockFinalized periodically tries to commit public rand until success or the block is finalized
+// error will be returned if maximum retries have been reached or the query to the consumer chain fails
+func (fp *FinalityProviderInstance) retryCommitPubRandUntilBlockFinalized(targetBlock *types.BlockInfo) (*types.TxResponse, error) {
+	var failedCycles uint32
+
+	// we break the for loop if the block is finalized or the public rand is successfully committed
+	// error will be returned if maximum retries have been reached or the query to the consumer chain fails
+	for {
+		// error will be returned if max retries have been reached
+		res, err := fp.CommitPubRand(targetBlock.Height)
```
gitferry, line 595: Maybe it's too much to put the whole `fp.CommitPubRand` here as we only need to retry the communication part, ie. `fp.cc.CommitPubRandList`.

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+	// TODO: should consider an implementation that deterministically create
+	//  randomness without saving it
+	pubRandList, err := fp.createPubRandList(startHeight, fp.cfg.NumPubRand)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate randomness: %w", err)
+	}
+	numPubRand := uint64(len(pubRandList))
+
+	// generate commitment and proof for each public randomness
+	commitment, proofList := types.CommitPubRandList(pubRandList)
```
gitferry, line 681:
```suggestion
	commitment, proofList := types.GetPubRandCommit(pubRandList)
```

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+		return nil, fmt.Errorf("failed to save public randomness to DB: %w", err)
+	}
+
+	// sign the commitment
+	schnorrSig, err := fp.signPubRandCommit(startHeight, numPubRand, commitment)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign the Schnorr signature: %w", err)
+	}
+
+	res, err := fp.cc.CommitPubRandList(fp.GetBtcPk(), startHeight, numPubRand, commitment, schnorrSig)
```
gitferry, line 694: So committing pub rand going after the `AddPubRandProofList`. What if the former fails and we retry, then the later will fail because of duplicate error.

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
 // SubmitFinalitySignature builds and sends a finality signature over the given block to the consumer chain
 func (fp *FinalityProviderInstance) SubmitFinalitySignature(b *types.BlockInfo) (*types.TxResponse, error) {
-	eotsSig, err := fp.signEotsSig(b)
+	sig, err := fp.signFinalitySig(b)
 	if err != nil {
 		return nil, err
 	}
 
+	// get public randomness at the height
+	prList, err := fp.createPubRandList(b.Height, 1)
```
gitferry, line 716: Maybe it's clearer to have an adaptor method like `getPubRandList(b.Height, 1)`

finality-provider/service/fp_instance.go
```diff
@@ -501,17 +748,29 @@ func (fp *FinalityProviderInstance) SubmitBatchFinalitySignatures(blocks []*type
 		return nil, fmt.Errorf("should not submit batch finality signature with zero block")
 	}
 
-	sigs := make([]*btcec.ModNScalar, 0, len(blocks))
+	// get public randomness list
+	prList, err := fp.createPubRandList(blocks[0].Height, uint64(len(blocks)))
```
gitferry, line 752: same here

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,150 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate randomness: %w", err)
+	}
+	numPubRand := uint64(len(pubRandList))
+
+	// generate commitment and proof for each public randomness
+	commitment, proofList := types.CommitPubRandList(pubRandList)
+
+	// store them to database
+	if err := fp.pubRandState.AddPubRandProofList(pubRandList, proofList); err != nil {
```
gitferry, line 684: Not sure we should return an error if the error is `ErrDuplicatePubRand` as it could be inconsistency between the remote committed pub rand and ones in the store

finality-provider/store/pub_rand.go
```diff
@@ -0,0 +1,136 @@
...
+	// mapping: pub_rand -> proof
+	pubRandProofBucketName = []byte("pub_rand_proof")
+)
+
+type PubRandProofStore struct {
+	db kvdb.Backend
+}
+
+// NewPubRandProofStore returns a new store backed by db
+func NewPubRandProofStore(db kvdb.Backend) (*PubRandProofStore, error) {
```
gitferry, line 21: Would we prune the store after the pub rand proof is useless?

gitferry (commented):
Nice work! It's difficult to get everything in place. Left some comments and some of them can be addressed in a separate PR

finality-provider/service/app.go
SebastianElvis: Is this already checked at https://github.com/babylonchain/finality-provider/blob/812d3b753585021e9f07b6ae5260da94466066a3/finality-provider/service/app.go#L357-L371?

finality-provider/service/fp_instance.go
SebastianElvis: Checked the code and we are already doing retry. Removed the TODO :v:

clientcontroller/babylon.go
```diff
@@ -110,55 +109,94 @@ func (bc *BabylonController) reliablySendMsgs(msgs []sdk.Msg, expectedErrs []*sd
...
+	fpPk *btcec.PublicKey,
+	block *types.BlockInfo,
+	pubRand *btcec.FieldVal,
+	proof []byte,
+	sig *btcec.ModNScalar,
+) (*types.TxResponse, error) {
+	cmtProof := cmtcrypto.Proof{}
+	if err := cmtProof.Unmarshal(proof); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
SebastianElvis: Suggest we think about it after we have implemented some optimisations. Optimisations (e.g., pollarding) could affect the proof format. Wdyt?

clientcontroller/interface.go
```diff
@@ -59,9 +67,6 @@ type ClientController interface {
 	// error will be returned if the consumer chain has not been activated
 	QueryActivatedHeight() (uint64, error)
 
-	// QueryLastFinalizedEpoch returns the last finalised epoch of Babylon
-	QueryLastFinalizedEpoch() (uint64, error)
```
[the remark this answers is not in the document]
SebastianElvis: Yes

finality-provider/service/eots_manager_adapter.go
SebastianElvis: Not really, as `HashToSign` only uses the three fields. We might need to modify Babylon side to make `HashToSign` a public function taking the three fields. The current usage is confusing indeed

clientcontroller/babylon.go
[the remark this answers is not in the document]
gitferry: No problem. Let's add a TODO to track this

finality-provider/service/eots_manager_adapter.go
SebastianElvis: Similarly, other fields are not used in `MsgToSign`. To avoid the confusion I created private functions to replace `HashToSign` and `MsgToSign`

finality-provider/store/pub_rand.go
SebastianElvis: This needs some thinking. The storage is held by the FP manager which manages multiple FPs, so we need something like `(fp PK || height) -> (pubRand, proof)`. 

> It could be helpful for range query and get the last committed height in case we need them in the future

From what I see in `ReadBucket` interface of DB, it only supports `ForEach` but not range queries. Also I remember last committed height is queried from Babylon rather than queried from local DB?

finality-provider/service/pub_rand_store_adapter.go
SebastianElvis: I followed `fpState`'s [implementation](https://github.com/babylonchain/finality-provider/blob/812d3b753585021e9f07b6ae5260da94466066a3/finality-provider/service/fp_store_adapter.go#L61-L65) tbh. Maybe you guys have done some research and found that the DB we are using is not thread safe?

finality-provider/service/fp_instance.go
SebastianElvis: This seems to be from some prior implementations. Agreed that retrying a function that already does retrying is too much. Maybe L594-615 should be moved outside the for loop then?

Also I suggest we do this in another PR -- this seems less relevant to the randomness commitment and this PR is big enough

finality-provider/service/fp_instance.go
SebastianElvis: Good point! Probably we need to break down the function a bit then, e.g., we first ensure pub rand commitment and proofs are generated locally, then we try to commit to Babylon

finality-provider/service/fp_instance.go
SebastianElvis: I thought about this as well and ended up not doing it, as the implmentation will be same as `createPubRandList` :sweat:

finality-provider/service/pub_rand_store_adapter.go
gitferry: `fpState` needs `Mutex` because it has an in-memory cache `store.StoredFinalityProvider`. But `pubRandState` does not need it, correct?

finality-provider/store/pub_rand.go
gitferry: It's true. Not sure we need it for now. I was thinking maybe in the future we want to know `{pubRand, proof}` by height. Currently, we can get it by asking the `eots-manager` to get the `pubRand` by height and then the proof, right? If so, maybe using `pubRand` as key is good for now

finality-provider/service/fp_instance.go
gitferry: Yeah, the prior implementation also has this issue. The issue is not retrying a function that already does retrying. It is that retrying a function that contains stuff that should not be retried. We should only retry `fp.cc.CommitPubRandList`. Retrying other stuff might cause problems, e.g., storing duplicate pub rand.

finality-provider/service/fp_instance.go
gitferry: I think the problem is that generating the commit is also part of the retry. We should only retry `fp.cc.CommitPubRandList`. See prior comments https://github.com/babylonchain/finality-provider/pull/354#discussion_r1618122214

finality-provider/service/fp_instance.go
gitferry: I think from fp's perspective, it's `getPubRandList` but from eots-manager's perspective, there's not difference between `get` or `create` as it does not store pub rand

SebastianElvis: Let's rename it to `getPubRandList` then. FP retrieves the public randomness from EOTS mgr rather than asking it to create it anyway

finality-provider/service/fp_instance.go
SebastianElvis: Right, then this needs some major refactoring on the `CommitPubRand` function. For now I avoided returning error upon storing duplicated proofs and added a TODO. A proper refactoring should be separating the process of 1) determining the starting height to commit, 2) generating pub rand and inclusion proofs, and 3) committing public randomness. Only 3) should be a part of `retryCommitPubRandUntilBlockFinalized`. Also the flow of `retryCommitPubRandUntilBlockFinalized` might need to change a bit -- imo committing public randomness needs to be under `select` statement as well, and we terminate either upon a block is finalised or pub rand is successfully committed.

finality-provider/store/pub_rand.go
SebastianElvis: There is a [TODO](https://github.com/babylonchain/finality-provider/blob/812d3b753585021e9f07b6ae5260da94466066a3/finality-provider/store/pub_rand.go#L136-L137) for it. I didn't do it as I'm not fully sure about when to remove proofs. Also if one day we want to store the whole Merkle tree rather than each proof for storage optimisation, then removing becomes tricky as removing certain branches will affect proofs of subsequent pub rand.

finality-provider/service/pub_rand_store_adapter.go
SebastianElvis: Hmm I see. removed the lock then

SebastianElvis:
Thanks @gitferry @RafilxTenfen for the comments! I have addressed all of them, mostly about handling duplicated pub rand proof storage, some wrapper functions and removing some unused code. Please take a look again

finality-provider/service/fp_instance.go
```diff
@@ -501,17 +753,29 @@ func (fp *FinalityProviderInstance) SubmitBatchFinalitySignatures(blocks []*type
...
 
-	sigs := make([]*btcec.ModNScalar, 0, len(blocks))
+	// get public randomness list
+	prList, err := fp.getPubRandList(blocks[0].Height, uint64(len(blocks)))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get public randomness list: %v", err)
+	}
+	// get proof list
+	proofBytesList, err := fp.pubRandState.GetPubRandProofList(prList)
+	if err != nil {
```
KonradStaniec, line 763: this can return error if even one of the randomness in the list does not have created proof. I wonder when it can happen and how we should handle this error ? 

From what I see, this error will bubble up and will be converted to criticial error, which from what I understand will never be recovered.

Maybe in case of this error we should try o recreate proofs for given heights ?

KonradStaniec (approved):
no blockers from side at this point but we must test it on some devnet in a lot of failure scenarios.

More generally, the code is getting quite complex and is getting hard to comprehend all failure modes.

Also we may strongly consider avoiding local databse to make code a bit more resilient (at price of higher memory usage)

finality-provider/service/fp_instance.go
gitferry: Sounds good 👍 Let's resolve it in a separate PR and have tests covering possible cases

finality-provider/service/pub_rand_store_adapter.go
gitferry, line 1: Does it still make sense to have `pubRandState`?

gitferry (approved):
LGTM!

finality-provider/service/app.go
RafilxTenfen: We checked for the FPD key not the eots one, but thinking better this is not related to this PR, I will open an issue for this

RafilxTenfen (approved):
:tada:

finality-provider/service/app.go
RafilxTenfen: issue created at https://github.com/babylonchain/finality-provider/issues/361

clientcontroller/babylon.go
```diff
@@ -227,34 +278,32 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
...
+func (bc *BabylonController) QueryLastCommittedPublicRand(fpPk *btcec.PublicKey, count uint64) (map[uint64]*finalitytypes.PubRandCommitResponse, error) {
+	fpBtcPk := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
+
+	pagination := &sdkquery.PageRequest{
+		// NOTE: the count is limited by pagination queries
+		Limit:   count,
+		Reverse: true,
+	}
+
+	res, err := bc.bbnClient.QueryClient.ListPubRandCommit(fpBtcPk.MarshalHex(), pagination)
```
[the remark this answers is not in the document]
SebastianElvis: The only purpose for this query is to get the last commited height, i.e., the last height covered by the last commitment. It is a useful query (FP needs some way to know where to start committing public randomness) so maybe let's keep it?

finality-provider/service/fp_instance.go
```diff
@@ -471,15 +583,155 @@ func (fp *FinalityProviderInstance) checkBlockFinalization(height uint64) (bool,
...
+		return nil, fmt.Errorf("failed to generate randomness: %w", err)
+	}
+	numPubRand := uint64(len(pubRandList))
+
+	// generate commitment and proof for each public randomness
+	commitment, proofList := types.GetPubRandCommitAndProofs(pubRandList)
+
+	// store them to database
+	if err := fp.pubRandState.AddPubRandProofList(pubRandList, proofList); err != nil {
+		return nil, fmt.Errorf("failed to save public randomness to DB: %w", err)
```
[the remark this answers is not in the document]
SebastianElvis: Agree that having another DB is not optimal. However, getting rid of the proof storage needs some careful designs to manage and manipulate Merkle tree in memory to get right. In particular,

- We cannot regenerate Merkle tree upon every height. This consumes a lot of memory and computation overhead.
- We need to generate a Merkle tree in memory upon a certain height, and keep this Merkle tree in memory at a certain range of heights. This probably needs some breaking changes in the FP instance.
- We need to have some bootstrapping/crash recovery mechanism such that upon restarting FP, it regenerates the right Merkle tree and continue submitting finality signatures.
- The whole design is affected by the possible optimisations we will have.

This needs at least an ADR, or needs to be a part of the ADR for optimising Merkle tree randomness commitment. I will create a Github issue for this.

finality-provider/service/fp_instance.go
SebastianElvis: The error will occur only when the DB is corrupted for some reason (e.g., proof DB is removed for some reason). This is a valid concern and I added a TODO in the code. This is also another argument on keeping Merkle tree in memory actually.

finality-provider/service/pub_rand_store_adapter.go
SebastianElvis: We still need it to keep the proof in DB at the moment, right?

PR #360, opened 2024-05-28
feat: eots-manager add eots verifier

lesterli:
## Summary

To add one gRPC API `VerifyFinalitySig` to support EOTS verification for one rollup block

Run `eotsd start --verifier` to enable verifier gRPC server.

TODO:
- [ ] integration test with op-node, babylon node, eots aggregator


SebastianElvis:
- What's the relationship between this EOTS verifier and the EOTS verifier at https://github.com/babylonchain/babylon-evm-contracts/pull/8? I was under the impression that verification only takes place at EOTS aggregator and L1/L2 contract
- Why is EOTS manager providing this functionality? The EOTS manager here aims to manage the Schnorr/EOTS secret keys and secret randomness, rather than being a RPC service for verifying EOTS signatures

lesterli:
1. It is considered an off-chain verifier that any rollup users can run locally to verify the EOTS signatures themselves to achieve fast confirmation. And the contract in `babylon-evm-contracts` is the on-chain verifier.
2. We originally thought to use the EOTSManager codebase to reuse some code. If it is confused, we can split it as a separate daemon. Wdyt?

SebastianElvis:
> 1. It is considered an off-chain verifier that any rollup users can run locally to verify the EOTS signatures themselves to achieve fast confirmation. And the contract in `babylon-evm-contracts` is the on-chain verifier.

I see. Correct me if I'm wrong, this verifier will be a part of the EOTS aggregator?

> 2. We originally thought to use the EOTSManager codebase to reuse some code. If it is confused, we can split it as a separate daemon. Wdyt?

Yep let's move the functionality to another daemon program. In fact, I feel it might not be a part of the finality-provider repository as this targets different groups of users. If it's used in EOTS aggregator, how about we put it in the repository for EOTS aggregator?

lesterli:
> I see. Correct me if I'm wrong, this verifier will be a part of the EOTS aggregator?

In our design, it doesn't do the verification in the EOTS aggregator.

> If it's used in EOTS aggregator, how about we put it in the repository for EOTS aggregator?

The EOTS aggregator doesn't use it, but I think we can put this gRPC in the EOTS aggregator repo. 

@bap2pecs Wdyt?

bap2pecs:
> In fact, I feel it might not be a part of the finality-provider repository as this targets different groups of users

@SebastianElvis I think this part makes sense. so it will be confusing if the user of the chain has to run a program in the FP repo. I am in for the move as long as we don't need to share lots of code from this FP repo.

I am fine to create a new repo `finality-verifier` or just another daemon program within the same repo

lesterli:
> I think we can put this gRPC in the EOTS aggregator repo.

Correct it. 

Their target user is different, especially if the aggregator functions as a P2P network.

> I am fine to create a new repo `finality-verifier` or just another daemon program within the same repo

It makes sense.

SebastianElvis:
Yep having a new daemon for EOTS verifier sounds good to me. @gitferry wdyt?

bap2pecs:
rollup users will be running EOTS verifier

EOTS aggregator is run by a diff group of people

but for you q offline, EOTS aggregator will still verify the EOTS before storing in its database

gitferry:
Had an offline discussion with @SebastianElvis. A separate daemon for `eots-verifier` under the `finality-provider` repo sounds good to me 👍

eots-verifier/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+const (
+	defaultLogLevel       = "debug"
+	defaultLogFilename    = "eots-verifier.log"
+	defaultConfigFileName = "eots-verifier.conf"
+
+	defaultRpcListener   = "127.0.0.1:9528"
+	defaultBabylonRPC    = "http://127.0.0.1:26657"
+	defaultEotsAggRPC    = "http://127.0.0.1:9527"
+	defaultRollupRPC     = "http://127.0.0.1:8545"
+	defaultRollupChainId = "42069"
```
bap2pecs, line 22: called rollup is too specific. we should call it `defaultConsumerChainRPC`

bap2pecs: also default values should be sth that can work most of the time

but it's almost certain that the last 4 values should be changed when we run the daemon. so we should just remove those and only rely on the config file

eots-verifier/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+	DefaultDir = btcutil.AppDataDir("eots-verifier", false)
+)
+
+// Config is used to configure the EOTS verifier
+type Config struct {
+	LogLevel      string `long:"loglevel" description:"logging level"`
+	RpcListener   string `long:"rpclistener" description:"the listener for RPC connections, e.g., 127.0.0.1:1234"`
+	BabylonRPC    string `long:"babylon_rpc" description:"connect to the Babylon RPC service"`
+	EotsAggRPC    string `long:"eots_agg_rpc" description:"connect to the EOTS Aggregator RPC service"`
+	RollupRPC     string `long:"rollup_rpc" description:"connect to the Rollup RPC service"`
```
bap2pecs, line 39: Rollup -> consumer

eots-verifier/proto/eots_verifier.proto
```diff
@@ -0,0 +1,21 @@
...
+
+service EOTSVerifier {
+  // VerifyFinalitySig verifies that the finality signature is valid
+  rpc VerifyFinalitySig (VerifyFinalitySigRequest)
+      returns (VerifyFinalitySigResponse); 
+}
+
+message VerifyFinalitySigRequest {
+    // block_number is the block which need to be verified
+    uint64 block_number = 1;
```
bap2pecs, line 15: per my comments below, we need to also pass in the tx hash to be verified

also per design 3 in https://www.notion.so/snapchain/Security-Requirements-1076b96c44144e82a61b4ed393364399, we might ask FP to sign a range of blocks. so instead of verify

eots-verifier/service/server.go
```diff
@@ -0,0 +1,269 @@
...
+	wg.Wait()
+
+	return nil
+}
+
+// VerifyFinalitySig verifies that the finality signature is valid
+func (s *Server) VerifyFinalitySig(ctx context.Context, req *proto.VerifyFinalitySigRequest) (
+	*proto.VerifyFinalitySigResponse, error) {
+	height := req.BlockNumber
+	l2Block, err := s.rollupClient.BlockByNumber(ctx, new(big.Int).SetUint64(height))
```
bap2pecs, line 178: i think we shouldn't trust getting the data from the RPC. because we can't guarantee the RPC returns the same block that contains the same tx as the block the user accepts

let's think of the goal of the verifier program: to help a user to build up confidence that a tx he observes is in a block that has enough valid EOTS sigs

so the minimum input should be the tx hash. it has to be something that doesn't allow you to change this tx. block number itself doesn't provide the guarantee.

the flow should be:
- user obtains the tx hash
- user queries the RPC to get the block that contains the tx
- user gets all data (e.g. tx list) from the RPC
- user gets the signatures from the aggregator
- user gets the FP VP ...
- ...
- user verifies the EOTS signatures


add block number can ensure more security: to help a user to build up confidence that a tx he observes is in a **specific** block that has enough valid EOTS sigs

the flow should be:
- user obtains the tx hash
- user queries the RPC to get the block that contains the tx
- user check if the block number matches
- user gets all data (e.g. tx list) from the RPC
- user gets the signatures from the aggregator
- user gets the FP VP ...
- ...
- user verifies the EOTS signatures

eots-verifier/service/server.go
```diff
@@ -0,0 +1,269 @@
...
+
+	totalPower := uint64(0)
+	for _, fp := range bbnRes.FinalityProviders {
+		totalPower += fp.VotingPower
+	}
+
+	return bbnRes.FinalityProviders, totalPower, nil
+}
+
+func (s *Server) calculateValidPower(ctx context.Context, msg []byte, fpList []*bsctypes.FinalityProviderResponse, eotsInfos []*eotsservice.EOTSInfoResponse) (uint64, error) {
```
bap2pecs, line 238: calculateTotalPowerFromEotsVerifiedFps

eots-verifier/service/server.go
```diff
@@ -0,0 +1,269 @@
...
+	validPower := uint64(0)
+	for _, fp := range fpList {
+		for _, info := range eotsInfos {
```
bap2pecs, line 241: why are there two levels of for loops?

bap2pecs (changes requested):

bap2pecs:
> Had an offline discussion with @SebastianElvis. A separate daemon for `eots-verifier` under the `finality-provider` repo sounds good to me 👍

per my conversation w @SebastianElvis recently, it's probably better to put this in the https://github.com/babylonchain/babylon-da-sdk repo, making it kind of like a demo case for using the SDK as well

If there is some code sharing needed, we can always import from this FP repo

SebastianElvis:
Yes, putting the EOTS verifier under a new folder of the `babylon-da-sdk` repo seems to be a proper way

PR #364, opened 2024-06-04
e2e: wasmd setup

gusin13:
### Description
The wasmd setup had given me bit of pain, the pr is finally ready now. This pr sets up wasmd node in e2e tests and creates a basic test to upload the contract on wasmd consumer chain. 

### Note
In the future pr's I'll implement wasmd consumer interface so fp can submit sigs to the contract.


itest/wasmd_handler.go
```diff
@@ -0,0 +1,335 @@
...
+	Events []struct {
+		Type       string `json:"type"`
+		Attributes []struct {
+			Key   string `json:"key"`
+			Value string `json:"value"`
+		} `json:"attributes"`
+	} `json:"events"`
+}
+
+func (w *WasmdNodeHandler) StoreWasmCode(wasmFile string) (string, string, error) {
```
gusin13, line 228: just want to confirm my approach here, are there any readily available clients which submit the tx to wasmd chain? Like [this](https://github.com/CosmWasm/wasmd/blob/21b048d54e395ff9168e5c3037356a73797500ba/x/wasm/ibctesting/wasm.go#L53) one but its only for the test chain. I am concerned for all the operations needed to be performed on chain would I need to create such `exec.Command ` wrapper functions?

itest/test_manager.go
```diff
@@ -57,6 +57,7 @@ type TestManager struct {
 	Wg                sync.WaitGroup
 	BabylonHandler    *BabylonNodeHandler
 	EOTSServerHandler *EOTSServerHandler
+	WasmdHandler      *WasmdNodeHandler
```
SebastianElvis, line 60: `TestManager` currently focuses on testing Babylon FPs but not consumer FPs. How about we have 

```go
type ConsumerTestManager struct {
    *TestManager
    WasmdHandler      *WasmdNodeHandler
}
```

and add consumer functionalities on top of that?

itest/wasmd_handler.go
```diff
@@ -0,0 +1,335 @@
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	cmd.Stderr = os.Stderr
+
+	err := cmd.Run()
+	if err != nil {
+		return "", "", fmt.Errorf("error running wasmd store command: %v", err)
+	}
+
+	var txResp TxResponse
```
SebastianElvis, line 242: I remember this struct exists in Cosmos SDK or CometBFT?

itest/wasmd_handler.go
SebastianElvis: > I am concerned for all the operations needed to be performed on chain would I need to create such exec.Command wrapper functions?

I did not know any wrapper client for x/wasm module to my knoweldge (cc @maurolacy). Wrapping each tx submission in a function seems okay to me as at the end of the day you need to manually wire up these message anyway. What we need to write include 1) storing contract code, 2) instantiate contract code, and 3) some queries. Wdyt?

itest/wasmd_handler.go
```diff
@@ -0,0 +1,335 @@
...
+	return height, nil
+}
+
+type ListResponse struct {
+	CodeInfos []CodeInfo `json:"code_infos"`
+}
+
+type CodeInfo struct {
+	CodeID string `json:"code_id"`
+}
```
SebastianElvis, line 313: These seem already defined in `x/wasm` ([ref](https://github.com/CosmWasm/wasmd/blob/21b048d54e395ff9168e5c3037356a73797500ba/x/wasm/types/query.pb.go#L770-L776)). Can we use them directly?

SebastianElvis (approved):
Lgtm :+1:

Makefile
```diff
@@ -58,8 +59,15 @@ build-docker:
...
 
-test-e2e:
-	cd $(TOOLS_DIR); go install -trimpath $(BABYLON_PKG)
+install-babylond:
+	cd $(TOOLS_DIR); \
+	go install -trimpath $(BABYLON_PKG)
+
+install-wasmd:
+	cd $(TOOLS_DIR); \
+	go install -trimpath $(WASMD_PKG)
```
maurolacy, line 68: 👍🏼 Nice.

itest/consumer_e2e_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestConsumerStoreContract stores a contract in the consumer chain
+func TestConsumerStoreContract(t *testing.T) {
+	tm, _, _ := StartManagerWithFinalityProvider(t, 1)
+	defer tm.Stop(t)
+	// Store the Babylon contract in the consumer chain
+	babylonContractPath := "wasmd_contracts/babylon_contract.wasm"
```
maurolacy, line 17: Why don't use the standard / usual dir (`bytecode/`) here?

Also, why is this under a different folder struct (`itest/`)? Just curious.

itest/wasmd_handler.go
maurolacy: It's OK with me as well. Take a look at how `TestChain` is handling this in the ref. that you posted? (i.e. the `SendMsgs()` function).

You can then write your own equivalent wrapper functions.

maurolacy (approved):
LGTM. Thanks for doing this.

I still don't fully understand how this is being encapsulated, but it looks like a step in the right direction.

itest/wasmd_handler.go
```diff
@@ -0,0 +1,335 @@
...
+	TxHash string `json:"txhash"`
+	Events []struct {
+		Type       string `json:"type"`
+		Attributes []struct {
+			Key   string `json:"key"`
+			Value string `json:"value"`
+		} `json:"attributes"`
+	} `json:"events"`
+}
+
```
bap2pecs, line 227: this is different than the type used in the `ConsumerController` interface

```
import (
	"github.com/cosmos/relayer/v2/relayer/provider"
)

type TxResponse struct {
	TxHash string
	Events []provider.RelayerEvent
}
```

does this imply we should change the type in interface.go?

itest/wasmd_handler.go
lesterli: Indeed, the Cosmos SDK defines message [TxResponse](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/base/abci/v1beta1/abci.proto#L15) in the file `abci.proto`, and this is the generated struct [TxResponse](https://github.com/cosmos/cosmos-sdk/blob/main/types/abci.pb.go#L34).

itest/consumer_e2e_test.go
gusin13: i'll rename to `bytecode/`, but idk about `itest/` as this dir exists before this pr (both btc-staker and finality-provider repo follow this)

itest/wasmd_handler.go
gusin13: hmm interesting, i have removed the txresponse from my pr as suggested in this [comment](https://github.com/babylonchain/finality-provider/pull/364#discussion_r1628040066) but about the TxResponse that is in `types/` [here](https://github.com/babylonchain/finality-provider/blob/37429abec0a514c4dbf95074fb231ff8464fdca8/types/txresponse.go#L7) I am not sure. 

@SebastianElvis I checked this struct is used in babylon query client [ref](https://github.com/babylonchain/babylon-private/blob/ae2182029020b8873173db826f0b71eb2d3fd01a/client/client/tx.go#L66), any idea why are we using `*pv.RelayerTxResponse` here instead of [sdk.TxResponse](https://github.com/cosmos/cosmos-sdk/blob/80ba18e39e21bf1c324cc39e54cf6ae2ab2364a8/proto/cosmos/base/abci/v1beta1/abci.proto#L15)?

gusin13:
Note: lint and tests are failing in https://github.com/babylonchain/finality-provider/commits/base/consumer-chain-support/
after the recent rebase, will wait to merge this pr till those are fixed.

itest/wasmd_handler.go
gusin13: if i use the structs from cometbft/sdk i am getting unmarshall err as the types don't match. The cli cmd gives results in strings but the structs from cometbft/sdk expects things in int. 

`consumer_test_manager.go:40: failed to get latest block height from wasmd error unmarshalling status response: json: cannot unmarshal string into Go struct field ProtocolVersion.node_info.protocol_version.p2p of type uint64`

For now I am using the custom structs i made to unmarshal, have created an issue to explore this later and also added todo's
https://github.com/babylonchain/finality-provider/issues/366

itest/wasmd_handler.go
SebastianElvis: This is because the client package in Babylon is based on the client implemented in Go relayer https://github.com/cosmos/relayer/tree/main/client

bap2pecs: @SebastianElvis so do you think we need to change this as well to support CosmWasm contracts?

SebastianElvis: > so do you think we need to change this as well to support CosmWasm contracts?

For the client to the DA cosmwasm contract? Yep ideally we could have a generic client for accessing `x/wasm` module in https://github.com/CosmWasm/wasmd. @gusin13 will be working on that

PR #365, opened 2024-06-05
fix: fix lint/e2e error

gusin13:
Grpc/Protobuf dependencies were bumped in this commit
https://github.com/babylonchain/finality-provider/commit/b79ac3101a2ef8205b97a7d1a8a061fff792d03f#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R34-R35
due to which grps.Dial has become deprecated and causes lint failure. This pr replaces deprecated with grpc.NewClient

```
// Dial calls DialContext(context.Background(), target, opts...).
//
// Deprecated: use NewClient instead.  Will be supported throughout 1.x.
func Dial(target string, opts ...DialOption) (*ClientConn, error) {
	return DialContext(context.Background(), target, opts...)
}
```

Lint doesn't fail in dev branch as it still uses old version of grpc

### DIFF
--- a/eotsmanager/client/rpcclient.go
+++ b/eotsmanager/client/rpcclient.go
@@ -22,7 +22,7 @@ type EOTSManagerGRpcClient struct {
 }
 
 func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error) {
-	conn, err := grpc.Dial(remoteAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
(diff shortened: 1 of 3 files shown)


lesterli (approved):

bap2pecs:
seems `build_lint_test ` still fails?

gusin13:
yes I have fixed the lint but the tests are still failing, not able to find the root cause, still debugging..

SebastianElvis:
Pushed another commit to fix e2e

SebastianElvis (approved):

gusin13:
@SebastianElvis thanks, good stuff ❤️

PR #368, opened 2024-06-07
feat: wasmd clients

gusin13:
This pr sets up wasmd consumer client so fp app can directly submit finality sigs to the contract, currently the fp app will panic because the queries on contract side are unimplemented (for example poller requires information about finalized blocks but the smart contract doesn't have finality module implemented yet so it's not aware of which blocks are finalized)
I have added relevant TODO's due to the above, have also added e2e test for finality sig submission.


clientcontroller/wasmd_consumer.go
```diff
@@ -0,0 +1,313 @@
...
+	// empty response
+	return 0, nil
+}
+
+func (wc *WasmdConsumerController) QueryLatestBlockHeight() (uint64, error) {
+	// empty response
+	return 0, nil
+}
+
+func (wc *WasmdConsumerController) StoreWasmCode(wasmFile string) error {
```
bap2pecs, line 189: is this only supposed to be used in tests? if yes we can consider put it elsewhere or add a comment here

wasmdclient/config/wasmd_config.go
```diff
@@ -0,0 +1,96 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/cosmos/relayer/v2/relayer/chains/cosmos"
+)
+
+// WasmdConfig defines configuration for the Babylon client
+// adapted from https://github.com/strangelove-ventures/lens/blob/v0.5.1/client/config.go
+type WasmdConfig struct {
```
SebastianElvis, line 15: Probably all `wasmd` could be renamed to `wasm` in the `wasmdclient` package. All Cosmos SDK chains with CosmWasm support can use this client, not just the wasmd chain

finality-provider/config/wasmd.go
```diff
@@ -0,0 +1,61 @@
...
+		GasPrices:      "1ustake",
+		Debug:          true,
+		Timeout:        20 * time.Second,
+		BlockTimeout:   1 * time.Minute,
+		OutputFormat:   "direct",
+		SignModeStr:    "",
+	}
+}
+
+func WasmdConfigToQueryClientConfig(wc *WasmdConfig) *config.WasmdConfig {
```
SebastianElvis, line 44:
```suggestion
func (wc *WasmdConfig) ToQueryClientConfig *config.WasmdConfig {
```

finality-provider/config/config.go
```diff
@@ -81,6 +81,8 @@ type Config struct {
 
 	EVMConfig *EVMConfig `group:"evm" namespace:"evm"`
 
+	WasmdConfig *WasmdConfig `group:"wasmd" namespace:"wasmd"`
```
SebastianElvis, line 84: We might need a check that at most 1 of these two configs is not empty in `Validate`

clientcontroller/interface.go
```diff
@@ -108,6 +116,18 @@ func NewConsumerController(config *fpcfg.Config, logger *zap.Logger) (ConsumerCo
...
 			return nil, fmt.Errorf("failed to create EVM rpc client: %w", err)
 		}
+	case WasmdConsumerChainName:
+		tempApp := wasmapp.NewWasmApp(sdklogs.NewNopLogger(), dbm.NewMemDB(), nil, false, simtestutil.NewAppOptionsWithFlagHome(tempDir()), []wasmkeeper.Option{})
+		encodingConfig := wasmdparams.EncodingConfig{
+			InterfaceRegistry: tempApp.InterfaceRegistry(),
+			Codec:             tempApp.AppCodec(),
+			TxConfig:          tempApp.TxConfig(),
+			Amino:             tempApp.LegacyAmino(),
+		}
```
SebastianElvis, line 126: Could this be wrapped into a function?

clientcontroller/wasmd_consumer.go
```diff
@@ -0,0 +1,313 @@
...
+
+func (wc *WasmdConsumerController) Close() error {
+	if !wc.WasmdClient.IsRunning() {
+		return nil
+	}
+
+	return wc.WasmdClient.Stop()
+}
+
+var tempDir = func() string {
```
SebastianElvis, line 305: Is this used?

itest/consumer_test_manager.go
```diff
@@ -1,38 +1,111 @@
...
 )
 
 type ConsumerTestManager struct {
-	*TestManager
-	WasmdHandler *WasmdNodeHandler
+	BabylonHandler      *BabylonNodeHandler
+	FpConfig            *fpcfg.Config
+	BBNClient           *fpcc.BabylonController
+	WasmdHandler        *WasmdNodeHandler
+	WasmdConsumerClient *fpcc.WasmdConsumerController
```
SebastianElvis, line 35: Not for this PR: later would these two fields be generalised to sth like

```suggestion
	consumerHandler        ConsumerHandler
	consumerClient         fpcc.ConsumerController
```

where `ConsumerHandler` and `fpcc.ConsumerController` are interfaces?

finality-provider/config/wasmd.go
```diff
@@ -0,0 +1,61 @@
+package config
+
+import (
+	"time"
+
+	"github.com/babylonchain/finality-provider/wasmdclient/config"
+)
+
+type WasmdConfig struct {
```
SebastianElvis, line 9: We also need Babylon / BTC staking contract addresses here, such that the FP knows which contracts to invoke/query

itest/consumer_e2e_test.go
```diff
@@ -1,38 +1,205 @@
...
+//
+//	consumerChainID := "consumer-chain-test-1"
+//	_, err := ctm.BBNClient.RegisterConsumerChain(consumerChainID, "Consumer chain 1 (test)", "Test Consumer Chain 1")
+//	require.NoError(t, err)
+//
+//	ctm.CreateFinalityProvidersForChain(t, consumerChainID, 1)
+//}
+
+// TestSubmitFinalitySignature tests the finality signature submission to the btc staking contract using admin
+func TestSubmitFinalitySignature(t *testing.T) {
```
SebastianElvis, line 32: Given that we will have 

- Babylon e2e tests
- Babylon / EVM e2e tests
- Babylon / Wasm e2e tests

How about we put this test and other tests of wasm chain integrtaion under a test suite?

SebastianElvis:
Hmm from my local setup e2e fails with the following

```
➜  finality-provider git:(gusin13/consumer-clients) go test -timeout 300s -tags e2e -run ^TestSubmitFinalitySignature$ github.com/babylonchain/finality-provider/itest
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
2024/06/11 16:39:43 error stopping wasmd process: exit status 1
--- FAIL: TestSubmitFinalitySignature (73.56s)
    consumer_test_manager.go:107: Babylon node is started
    consumer_test_manager.go:118: Wasmd node is started
    consumer_test_manager.go:246: 
                Error Trace:    /home/runchao/Projects/Babylon/finality-provider/itest/consumer_test_manager.go:246
                                                        /home/runchao/Projects/Babylon/finality-provider/itest/consumer_e2e_test.go:106
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestSubmitFinalitySignature
FAIL
FAIL    github.com/babylonchain/finality-provider/itest 73.646s
FAIL
```

Looks like something fishy happens when shutting down the Babylon node?

SebastianElvis (approved):
Code change looks good to me :+1: CI failed though

clientcontroller/wasmd_consumer.go
gusin13: correct this is meant to be used only in tests, have added comments now.

clientcontroller/wasmd_consumer.go
gusin13: yes its used in 
```
		tempApp := wasmapp.NewWasmApp(sdklogs.NewNopLogger(), dbm.NewMemDB(), nil, false, simtestutil.NewAppOptionsWithFlagHome(tempDir()), []wasmkeeper.Option{})
		encodingConfig := wasmdparams.EncodingConfig{
			InterfaceRegistry: tempApp.InterfaceRegistry(),
			Codec:             tempApp.AppCodec(),
			TxConfig:          tempApp.TxConfig(),
			Amino:             tempApp.LegacyAmino(),
		}
```

wasmdclient/config/wasmd_config.go
gusin13: I see, sure I can rename but in the interface [here](https://github.com/babylonchain/finality-provider/pull/368/files#diff-3984e84857cffafeb06cb441c8c26264a80ed677435647f94a85f038511e1a63R120) each cosmos chain will need to have its entry as the encoding config will be different for each chain.

gusin13: i have renamed files/variables/functions to prefix `cosmwasm` as `wasm` is bit confusing. In short we have two naming conventions now, `cosmwasm` for generic cosmwasm enabled chains and `wasmd` which is used in e2e tests.

finality-provider/config/config.go
gusin13: if the configs of any of the consumer assigned in fp config (babylon, evm, cosmwasm chain) is invalid, the fp app won't start. 

i am not sure how to check if at most one of its valid because the consumer is decided in fp config (through chain name) and the validation process starts after this

itest/consumer_test_manager.go
gusin13: sure, actually i have renamed the files so its like this now

`WasmdConsumerClient *fpcc.CosmwasmConsumerController`

the controller is named to represent cosmwasm enabled chains, on the left side the variable is still WasmdConsumerClient because in the tests we are using wasmd as the consumer chain.

clientcontroller/interface.go
gusin13: sure, done

finality-provider/config/wasmd.go
gusin13: sure, good point, have added a todo for now as the fp app is not functional right now in e2e test, will do this once i fix this in next pr.

itest/consumer_e2e_test.go
gusin13: yep good idea, this requires some restructuring of test files, is it ok if i add a todo for now and then do this in next pr!

gusin13:
> Hmm from my local setup e2e fails with the following
> 
> ```
> ➜  finality-provider git:(gusin13/consumer-clients) go test -timeout 300s -tags e2e -run ^TestSubmitFinalitySignature$ github.com/babylonchain/finality-provider/itest
> service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
> service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
> 2024/06/11 16:39:43 error stopping wasmd process: exit status 1
> --- FAIL: TestSubmitFinalitySignature (73.56s)
>     consumer_test_manager.go:107: Babylon node is started
>     consumer_test_manager.go:118: Wasmd node is started
>     consumer_test_manager.go:246: 
>                 Error Trace:    /home/runchao/Projects/Babylon/finality-provider/itest/consumer_test_manager.go:246
>                                                         /home/runchao/Projects/Babylon/finality-provider/itest/consumer_e2e_test.go:106
>                 Error:          Received unexpected error:
>                                 exit status 1
>                 Test:           TestSubmitFinalitySignature
> FAIL
> FAIL    github.com/babylonchain/finality-provider/itest 73.646s
> FAIL
> ```
> 
> Looks like something fishy happens when shutting down the Babylon node?

build tags were missing when i added ci job passed, maybe its due to that

gusin13:
Thanks for the review guys, i have resolved most of the comments, for some others I have added Todo's as the pr is getting bit lengthy so will merge this now. I'll open another pr to fix pending things.

clientcontroller/interface.go
SebastianElvis: Not really, since this controller only uses wasm module but not other modules, they could share this encoding config

Issue #372, opened 2024-06-11
`failedCycles` does not count the case where public randomness is not successfully committed


gitferry:

gitferry:
addressed by #373

Issue #371, opened 2024-06-11
Finality provider instances can terminate if randomness is not committed before the block is finalized


gitferry:
In the current implementation, the randomness commitment loop will terminate if the block is finalized. In this case, submitting a finality signature will cause the program to terminate due to public randomness not found error.

gitferry:
addressed by #373

PR #373, opened 2024-06-12
chore: Fix issues in retry loops

gitferry:
This PR fixed two issues #371 and #372  related to the retry loops:
* `retryCheckRandomnessUntilBlockFinalized ` can be returned if the block is finalized but the randomness does not exist. In this case, we should skip this block for submitting a finality vote. Otherwise, the submission will fail with `ErrPubRandNotExist`. The PR fixed this issue by introducing a `bool` indicating whether the target block is finalized or not.
* within the `retryCheckRandomnessUntilBlockFinalized` loop, if the randomness is never submitted, the loop will stall because the maximum retry time is only applied in errors. This PR addressed this by also limiting the time of retries if the randomness does not exist.

This PR also added an error type `ErrFinalityProviderShutDown` to close the retry loops gracefully upon shutdown.

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -207,11 +207,18 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
 				continue
 			}
 			// check whether the randomness has been committed
-			// we should stall here until we have randomness committed at this
-			// height, otherwise, we might miss blocks
-			if err := fp.retryCheckRandomnessUntilBlockFinalized(b); err != nil {
-				fp.reportCriticalErr(err)
+			// the retry will end if max retry times is reached
+			// or the target block is finalized
+			isFinalized, err := fp.retryCheckRandomnessUntilBlockFinalized(b)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


finality-provider/service/fp_instance.go
```diff
@@ -207,19 +207,28 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+			isFinalized, err := fp.retryCheckRandomnessUntilBlockFinalized(b)
+			if err != nil {
+				if !errors.Is(err, ErrFinalityProviderShutDown) {
+					fp.reportCriticalErr(err)
+				}
 				break
 			}
+			// the block is finalized, no need to submit finality signature
+			if isFinalized {
+				continue
```
SebastianElvis, line 221: Do we need to call `MustSetLastProcessedHeight` here?

SebastianElvis (approved):

finality-provider/service/fp_instance.go
gitferry: Good question. Not sure if it will cause issues if we not increment the last processed height but I think we should

PR #377, opened 2024-06-15
refactor: update `TxResponse` with generic `Events`

lesterli:
This PR refactors `TxResponse` to make the field `Events` more generic.

```Go
type TxResponse struct {
	TxHash string
	Events []byte
}
```

### DIFF
--- a/clientcontroller/babylon_consumer.go
+++ b/clientcontroller/babylon_consumer.go
@@ -2,6 +2,7 @@ package clientcontroller
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -129,7 +130,7 @@ func (bc *BabylonConsumerController) CommitPubRandList(
 		return nil, err
 	}
 
-	return &types.TxResponse{TxHash: res.TxHash, Events: types.GetEventsBytes(res.Events)}, nil
+	return &types.TxResponse{TxHash: res.TxHash}, nil
 }
 
 // SubmitFinalitySig submits the finality signature via a MsgAddVote to Babylon
@@ -166,7 +167,7 @@ func (bc *BabylonConsumerController) SubmitFinalitySig(
 		return nil, err
 	}
 
-	return &types.TxResponse{TxHash: res.TxHash, Events: types.GetEventsBytes(res.Events)}, nil
+	return &types.TxResponse{TxHash: res.TxHash, Events: fromCosmosEventsToBytes(res.Events)}, nil
 }
 
 // SubmitBatchFinalitySigs submits a batch of finality signatures to Babylon
@@ -211,7 +212,7 @@ func (bc *BabylonConsumerController) SubmitBatchFinalitySigs(
 		return nil, err
 	}
 
-	return &types.TxResponse{TxHash: res.TxHash, Events: types.GetEventsBytes(res.Events)}, nil
+	return &types.TxResponse{TxHash: res.TxHash}, nil
 }
 
 // QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height
@@ -358,4 +359,12 @@ func (bc *BabylonConsumerController) Close() error {
 	}
 
 	return bc.bbnClient.Stop()
+}
+
+func fromCosmosEventsToBytes(events []provider.RelayerEvent) []byte {
+	bytes, err := json.Marshal(events)
+	if err != nil {
+		return nil
+	}
+	return bytes
 }
--- a/clientcontroller/babylon_consumer.go
+++ b/clientcontroller/babylon_consumer.go
@@ -107,7 +107,7 @@ func (bc *BabylonConsumerController) CommitPubRandList(
 	numPubRand uint64,
 	commitment []byte,
 	sig *schnorr.Signature,
-) (types.TxResponse, error) {
+) (*types.TxResponse, error) {
 	msg := &finalitytypes.MsgCommitPubRandList{
 		Signer:      bc.mustGetTxSigner(),
 		FpBtcPk:     bbntypes.NewBIP340PubKeyFromBTCPK(fpPk),
@@ -129,7 +129,7 @@ func (bc *BabylonConsumerController) CommitPubRandList(
 		return nil, err
 	}
 
-	return &types.BabylonTxResponse{TxHash: res.TxHash, Events: res.Events}, nil
+	return &types.TxResponse{TxHash: res.TxHash, Events: types.GetEventsBytes(res.Events)}, nil
 }
 
 // SubmitFinalitySig submits the finality signature via a MsgAddVote to Babylon
@@ -139,7 +139,7 @@ func (bc *BabylonConsumerController) SubmitFinalitySig(
 	pubRand *btcec.FieldVal,
 	proof []byte, // TODO: have a type for proof
 	sig *btcec.ModNScalar,
-) (types.TxResponse, error) {
+) (*types.TxResponse, error) {
 	cmtProof := cmtcrypto.Proof{}
 	if err := cmtProof.Unmarshal(proof); err != nil {
 		return nil, err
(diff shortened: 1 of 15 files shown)


bap2pecs:
we had some discussion here https://github.com/babylonchain/finality-provider/pull/364#discussion_r1630642797 but it seems we haven't decided about the response type in the interfaces

cc @SebastianElvis @gusin13 @gitferry

lesterli:
For both the controller and consumer controller, the interface just needs `TxHash` and `Events`, the [sdk.TxResponse](https://github.com/cosmos/cosmos-sdk/blob/main/types/abci.pb.go#L34) is complex, and not enough generic when one consumer controller needs to interact with a non-Cosmos chain.
I proposed to use an interface `TxResponse` as in this PR.

types/txresponse.go
```diff
@@ -1,10 +1,57 @@
...
-type TxResponse struct {
+// TxResponse is the interface that handles different types of transaction responses
+type TxResponse interface {
+	GetTxHash() string
+	GetEventsBytes() []byte
+}
+
+var _ TxResponse = &BabylonTxResponse{}
+
+type BabylonTxResponse struct {
```
SebastianElvis, line 17:
```suggestion
type CosmosTxResponse struct {
```

This response is generic to all Cosmos SDK chains. For example, other Cosmos SDK chains that integrate with Babylon via a Go module could also use this response.

SebastianElvis (approved):
Concept ACK and the code lgtm! let's wait for the comment from @gitferry as well

types/txresponse.go
lesterli: I prefer to update the [CosmwasmTxResponse](https://github.com/babylonchain/finality-provider/blob/refactor/tx-response/types/txresponse.go#L36-L45) with the below definition:
```Go
type CosmosTxResponse struct {
	TxHash string
	Events []struct {
		Type       string
		Attributes []struct {
			Key   string
			Value string
			Index bool
		}
	}
}
```
It is more generic to all Cosmos SDK chains b/c it is part of the [sdk.TxRespons](https://github.com/cosmos/cosmos-sdk/blob/main/types/abci.pb.go#L34).

SebastianElvis: `provider.RelayerEvent` can be converted to the event type in `sdk.TxResponse` right? Maybe we can do the conversion here?

types/txresponse.go
```diff
@@ -1,10 +1,57 @@
...
+	"encoding/json"
+
 	"github.com/cosmos/relayer/v2/relayer/provider"
 )
 
-type TxResponse struct {
+// TxResponse is the interface that handles different types of transaction responses
+type TxResponse interface {
+	GetTxHash() string
+	GetEventsBytes() []byte
```
gitferry, line 12: Is it used anywhere? I can see we need `TxResponse.Events in `TestSubmitFinalitySignatureAndExtractPrivKey`, which is for testing purposes only. I think this is the one that needs to be generalized other than `TxResponse` itself.

gitferry (commented):
Maybe I was missing something but I don't see making `TxResponse` an interface addresses the generalization issue

types/txresponse.go
lesterli: Yes, `Events` only is used in testing. 
So now the original `TxResponse` with the relayer events should be fine. I will close this PR.

SebastianElvis: I suppose we still need a generic interface that supports getting tx hash? if so this PR is still useful

SebastianElvis: > I can see we need TxResponse.Events in TestSubmitFinalitySignatureAndExtractPrivKey, which is for testing purposes only. I think this is the one that needs to be generalized other than TxResponse itself.

Hmm if that's the case then we could simply cast the response to the relayer one, and then finish all assertions in the test. This way we do not need `GetEventsBytes` anymore. Wdyt?

gitferry: Isn't a `string` field in `TxResponse` struct generic enough? I think the purpose of this PR is to make events type generic. If so, we should have a generic event type within `TxReponse` struct other than making `TxResponse` an interface because we already have `ConsumerController` as an interface

SebastianElvis: Hmm so you are suggesting to make `TxResponse` a struct rather than an interface. This also makes sense to me. The only thing I see is that the client could not assert `TxResponse` to concrete types that might contain more fields. No strong opinion on this

types/txresponse.go
lesterli: yes, we can convert it here.

types/txresponse.go
lesterli: Ok, we can have this PR. @gitferry Wdyt?

lesterli: Updated it based on comments. @gitferry @SebastianElvis

types/txresponse.go
```diff
@@ -1,10 +1,33 @@
...
+// Not every consumer has Events thing in their response,
+// so consumer client implementations need to care about Events field.
 type TxResponse struct {
 	TxHash string
-	Events []provider.RelayerEvent
+	// JSON-encoded data, now it is for testing purposes only
+	Events []byte
+}
+
+func GetEventsBytes(events []provider.RelayerEvent) []byte {
```
gitferry, line 19:
```suggestion
func FromCosmosEventsToBytes(events []provider.RelayerEvent) []byte {
```

types/txresponse.go
```diff
@@ -1,10 +1,33 @@
...
+	bytes, err := json.Marshal(events)
+	if err != nil {
+		return nil
+	}
+	return bytes
+}
+
+func (res *TxResponse) ParseEvents() ([]provider.RelayerEvent, error) {
```
gitferry, line 27:
```suggestion
func ParseCosmosEvents() ([]provider.RelayerEvent, error) {
```

gitferry (commented):
I think the encode/decode functions are better implemented in chain-specific packages other than in the `types/txresponse.go` which should remain as generic as possible. Maybe we can remove them to `babylon_consumer.go` as only Babylon chain uses it?

lesterli:
> Maybe we can remove them to `babylon_consumer.go` as only Babylon chain uses it?

Both `babylon.go` and `babylon_consumer.go` use it.

gitferry:
Moving to `babylon_consumer.go` makes more sense because we only need events from the response of `SubmitFinalityProvider`. In fact, for other responses, we don't need to set `Events`.

lesterli:
> In fact, for other responses, we don't need to set `Events`.

Got it, just updated the codes.

clientcontroller/babylon_consumer.go
```diff
@@ -359,3 +360,11 @@ func (bc *BabylonConsumerController) Close() error {
 
 	return bc.bbnClient.Stop()
 }
+
+func (bc *BabylonConsumerController) fromCosmosEventsToBytes(events []provider.RelayerEvent) []byte {
```
gitferry, line 364: why does it have to be a method of `BabylonConsumerController`?

finality-provider/service/fp_instance.go
```diff
@@ -841,6 +847,14 @@ func (fp *FinalityProviderInstance) TestSubmitFinalitySignatureAndExtractPrivKey
 	return res, privKey, nil
 }
 
+func (fp *FinalityProviderInstance) parseCosmosEvents(eventsData []byte) ([]provider.RelayerEvent, error) {
```
gitferry, line 850: why does it have to be a method of `BabylonConsumerController`? It seems the receiver is not used.

clientcontroller/babylon_consumer.go
```diff
@@ -211,7 +212,7 @@ func (bc *BabylonConsumerController) SubmitBatchFinalitySigs(
 		return nil, err
 	}
 
-	return &types.TxResponse{TxHash: res.TxHash, Events: res.Events}, nil
+	return &types.TxResponse{TxHash: res.TxHash, Events: bc.fromCosmosEventsToBytes(res.Events)}, nil
```
gitferry, line 215:
```suggestion
	return &types.TxResponse{TxHash: res.TxHash}, nil
```
This is not used in tests

clientcontroller/babylon_consumer.go
lesterli: Ok, only a helper function is fine.

gitferry (approved):
Thanks for all the changes. LGTM!

PR #379, opened 2024-06-18
feat: OPStack L2 consumer controller

lesterli:
## Summary

This PR implements the OPStack L2 consumer controller which sends messages to Babylon CosmWasm contract.

* Commits a list of Schnorr public randomness 
* Submits the finality signature


clientcontroller/interface.go
```diff
@@ -107,10 +107,10 @@ func NewConsumerController(config *fpcfg.Config, logger *zap.Logger) (ConsumerCo
 		if err != nil {
 			return nil, fmt.Errorf("failed to create Babylon rpc client: %w", err)
 		}
-	case EVMConsumerChainName:
-		ccc, err = NewEVMConsumerController(config.EVMConfig, logger)
+	case OPStackL2ConsumerChainName:
+		ccc, err = NewOPStackL2ConsumerController(config.BabylonConfig, config.OPStackL2Config, logger)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create EVM rpc client: %w", err)
+			return nil, fmt.Errorf("failed to create EVM consumer client: %w", err)
```
gitferry, line 113:
```suggestion
			return nil, fmt.Errorf("failed to create OPStack L2 client: %w", err)
```

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+		&bbnConfig,
+		logger,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Babylon client: %w", err)
+	}
+
+	keyRec, err := bbnClient.GetKeyring().Key(bbnConfig.Key)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get key address: %s", err))
```
gitferry, line 50: maybe we can return error and let the caller decide?

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// submitterAddress retrieves address based on key name which is configured in
+	// cfg *stakercfg.BBNConfig.
+	submitterAddress, err := keyRec.GetAddress()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get key address: %s", err))
```
gitferry, line 57: same here

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	keyRec, err := bbnClient.GetKeyring().Key(bbnConfig.Key)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get key address: %s", err))
+	}
+
+	// submitterAddress retrieves address based on key name which is configured in
+	// cfg *stakercfg.BBNConfig.
+	submitterAddress, err := keyRec.GetAddress()
```
gitferry, line 55: why do we need submitter address?

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+
+type OPStackL2ConsumerController struct {
+	bbnClient *bbnclient.Client
+	cfg       *fpcfg.OPStackL2Config
+	logger    *zap.Logger
+}
+
+func NewOPStackL2ConsumerController(
+	bbnCfg *fpcfg.BBNConfig,
+	opl2Cfg *fpcfg.OPStackL2Config,
```
gitferry, line 33: So, the OPStackL2 client will use the same rpc address configured in `BBNConfig`? The `GRPCAddress` in `OPStackL2Config` seems not used. I would suggest we extend `OPStackL2Config` by copying necessary fields from `BBNConfig` and have `BBNConfigToBabylonConfig` method to do the conversion.

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+	bbnConfig.SubmitterAddress = submitterAddress.String()
+
+	return &OPStackL2ConsumerController{
+		bbnClient,
+		opl2Cfg,
+		logger,
+	}, nil
+}
+
+func (ec *OPStackL2ConsumerController) ExecuteContract(contractAddress string, payload []byte) (*provider.RelayerTxResponse, error) {
```
gitferry, line 68: maybe we don't need `contractAddress` as a parameter as it can be obtained from `ec.cfg.OPFinalityGadgetAddress`

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	res, err := ec.ExecuteContract(ec.cfg.OPFinalityGadgetAddress, msgData)
+	if err != nil {
+		return nil, err
+	}
+	return &types.TxResponse{TxHash: res.TxHash}, nil
+}
+
+// SubmitBatchFinalitySigs submits a batch of finality signatures
+func (ec *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
```
gitferry, line 161: We don't need to implement it for OPStack L2?

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+func (ec *OPStackL2ConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.PublicKey, count uint64) (map[uint64]*finalitytypes.PubRandCommitResponse, error) {
+
+	return nil, nil
+}
+
+func (ec *OPStackL2ConsumerController) Close() error {
+	if err := ec.bbnClient.Stop(); err != nil {
+		return err
+	}
+	return nil
```
gitferry, line 272:
```suggestion
	return ec.bbnClient.Stop()
```

types/finality.go
gitferry, line 1: seems it is only needed for `OPStackL2`? If so, I was thinking maybe it's better to have a separate package for each consumer chain and put all the chain specific stuff there, e.g., types, configs, implementations

gitferry (commented):
One general comment is that maybe we should have a separate package under `clientcontroller` so that it can holds all the chain specific stuff e.g., types, configs, implementations. Wdyt? cc @SebastianElvis @bap2pecs

types/finality.go
bap2pecs: agree that this should either be renamed or moved to a separate paakge

clientcontroller/opstackl2_consumer.go
lesterli: OK

clientcontroller/opstackl2_consumer.go
lesterli: It's no longer needed, but I forgot to remove it. I previously used it to submit Cosmos Tx.

clientcontroller/opstackl2_consumer.go
lesterli: Yes, it uses the same RPC address, and the current design aims to avoid having FP make redundant configurations.

clientcontroller/opstackl2_consumer.go
lesterli: yes

clientcontroller/opstackl2_consumer.go
lesterli: We also need to implement it for the FP fast sync case.

types/finality.go
lesterli: Yes, it makes sense.

clientcontroller/opstackl2_consumer.go
gitferry: I wouldn't worry about redundancy but I think mixing configs across different consumer chains would be troublesome

lesterli: Ok, it makes sense, let's extend the `OPStackL2Config`.

types/finality.go
SebastianElvis: Agreed that the types and impl of consumer interface could be bundled into a separate package. However, the config might need to stay under `finality-provider/config/`. At the end of the day a OP stack L2 FP needs to specify something in the config file.

finality-provider/config/opstackl2.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import (
+	"fmt"
+)
+
+type OPStackL2Config struct {
+	GRPCAddress             string `long:"grpc-address" description:"address of the grpc server(e.g.Babylon) to connect to"`
```
SebastianElvis, line 8:
```suggestion
	GRPCAddress             string `long:"grpc-address" description:"address of the grpc server (e.g.Babylon) to connect to"`
```

clientcontroller/opstackl2_consumer.go
SebastianElvis: Agreed, it's probably not a good idea to let the controller to touch the `BBNConfig` object which is solely for Babylon FPs. Instead the `OPStackL2Config` object should contain the RPC address of the node that hosts the smart contract, and this controller should be able to bootstrap with `OPStackL2Config` only.

clientcontroller/opstackl2_consumer.go
```diff
@@ -0,0 +1,273 @@
...
+	cmtcrypto "github.com/cometbft/cometbft/proto/tendermint/crypto"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/relayer/v2/relayer/provider"
+	"go.uber.org/zap"
+)
+
+var _ ConsumerController = &OPStackL2ConsumerController{}
+
+type OPStackL2ConsumerController struct {
+	bbnClient *bbnclient.Client
```
SebastianElvis, line 26: Could we extend `cosmwasmclient/` and use that, rather than a Babylon client? Most functions under Babylon client could not be used in this controller and `cosmwasmclient/` provides a "just enough" client for submitting txs to CosmWasm smart contracts

SebastianElvis (commented):
Great work on fleshing out the consumer controller! Agree that we probably need some separation between consumer controller objects.

types/finality.go
gitferry: If OP stack L2 FP needs specific configs not related to client, we can add them under `finality-provider/config/`. Other than those, I think client specific configs should stay in specific consumer package.

Btw, why would an OP stack L2 FP needs special configs? Does it mean it will change the core logic?

SebastianElvis: In order to connect the FP to the data availability (DA) contract that stores all finality signatures, FP will at least need to know the RPC address of the Cosmos SDK node that hosts the contract, metadata needed for submitting txs to the Cosmos SDK node, and the address of the contract. These need to be specified via the config, right?

> Btw, why would an OP stack L2 FP needs special configs? Does it mean it will change the core logic?

Core logic will be the same. It's just that FP submits pub rand and finality sig in different message formats

gitferry: If it's communication related, why can it be in the consumer client config?

SebastianElvis: It's possible, but at the end of the day the consumer client config will be a part of the finality provider's config, no?

clientcontroller/opstackl2_consumer.go
lesterli: To create a `Client` under the `cosmwasmclient`, we need the [EncodingConfig](https://github.com/babylonchain/finality-provider/blob/base/consumer-chain-support/cosmwasmclient/client/client.go#L24), do you know how to get it for the Babylon chain?

gusin13: the encoding cfg for bbn is created here [ref1](https://github.com/babylonchain/finality-provider/blob/77aa7b5b57e3cdc664bf75847c5eb6d927709511/clientcontroller/babylon_consumer.go#L47) [ref2](https://github.com/babylonchain/babylon/blob/7ecdfdeefc3763001e601c289599c119ad647625/client/client/client.go#L59)

so iiuc you'd use cosmwasm client but use babylon's encoding cfg instead, i think there might be more changes involved then b/c then type of encoding cfg here is specific to wasmd [ref](https://github.com/babylonchain/finality-provider/blob/77aa7b5b57e3cdc664bf75847c5eb6d927709511/clientcontroller/cosmwasm_consumer.go#L37), we probably need a generic type here then 🤔

gusin13: or probably the same type would work wasmdparams.EncodingConfig, the internal things would be replaced with babylon specific sth like this

```
	case EVMConsumerChainName:
		encodingCfg := bbn.GetEncodingConfig()
		ccc, err = NewCosmwasmConsumerController(config.CosmwasmConfig, encodingCfg, logger)
		if err != nil {
			return nil, fmt.Errorf("failed to create Wasmd rpc client: %w", err)
		}
```

finality-provider/config/opstackl2.go
```diff
@@ -0,0 +1,60 @@
...
+	if cfg.OPFinalityGadgetAddress == "" {
+		return fmt.Errorf("the contract address of the op-finality-gadget not specified")
+	}
+	if cfg.RPCAddr == "" {
+		return fmt.Errorf("the rpc address of the node that hosts the smart contract not specified")
+	}
+	return nil
+}
+
+func OPStackL2ConfigToCosmwasmConfig(cfg *OPStackL2Config) cwcfg.CosmwasmConfig {
```
SebastianElvis, line 43:
```suggestion
func (cfg *OPStackL2Config) ToCosmwasmConfig() *cwcfg.CosmwasmConfig {
```

SebastianElvis (approved):
Great work! The refactoring  of `clientcontroller/` is very clean :+1: would be good to see comments from @gitferry as well

clientcontroller/opstackl2/consumer.go
```diff
@@ -0,0 +1,292 @@
...
+	wasmdEncodingCfg := wasmdparams.EncodingConfig{
+		InterfaceRegistry: bbnEncodingCfg.InterfaceRegistry,
+		Codec:             bbnEncodingCfg.Codec,
+		TxConfig:          bbnEncodingCfg.TxConfig,
+		Amino:             bbnEncodingCfg.Amino,
+	}
+
+	bbnClient, err := cwclient.New(
+		&cwConfig,
+		"Babylon",
```
gitferry, line 56: nit: maybe we can make it constant in `factory.go`?

finality-provider/config/opstackl2.go
```diff
@@ -0,0 +1,60 @@
...
+	GasPrices      string        `long:"gas-prices" description:"comma separated minimum gas prices to accept for transactions"`
+	KeyDirectory   string        `long:"key-dir" description:"directory to store keys in"`
+	Debug          bool          `long:"debug" description:"flag to print debug output"`
+	Timeout        time.Duration `long:"timeout" description:"client timeout when doing queries"`
+	BlockTimeout   time.Duration `long:"block-timeout" description:"block timeout when waiting for block events"`
+	OutputFormat   string        `long:"output-format" description:"default output when printint responses"`
+	SignModeStr    string        `long:"sign-mode" description:"sign mode to use"`
+}
+
+func (cfg *OPStackL2Config) Validate() error {
```
gitferry, line 30: we can copy the validation code from `CosmwasmConfig` here

gitferry (approved):
Thanks for the refactoring! It's super clean now. Great work! Some nits:

clientcontroller/opstackl2/consumer.go
lesterli: Ok

PR #381, opened 2024-06-18
feat: cosmwasm consumer queries

gusin13:
This pr

- implements cosmwasm consumer query interfaces 
- restructures files and functions

Note
- the fp daemon will still not work until `bcd` is integrated in tests, i will work on this in next pr. 
- the query related to pub randomness is not implemented on contract side, i will open a pr there and then integrate that here in next pr.


types/cosmwasm.go
lesterli, line 1: I think this https://github.com/babylonchain/finality-provider/pull/379#pullrequestreview-2124233586 also applies to the Cosmwasm consumer.

finality-provider/config/cosmwasm.go
```diff
@@ -7,20 +7,22 @@ import (
...
+	AccountPrefix             string        `long:"acc-prefix" description:"account prefix to use for addresses"`
+	KeyringBackend            string        `long:"keyring-type" description:"type of keyring to use"`
+	GasAdjustment             float64       `long:"gas-adjustment" description:"adjustment factor when using gas estimation"`
+	GasPrices                 string        `long:"gas-prices" description:"comma separated minimum gas prices to accept for transactions"`
+	KeyDirectory              string        `long:"key-dir" description:"directory to store keys in"`
+	Debug                     bool          `long:"debug" description:"flag to print debug output"`
+	Timeout                   time.Duration `long:"timeout" description:"client timeout when doing queries"`
+	BlockTimeout              time.Duration `long:"block-timeout" description:"block timeout when waiting for block events"`
+	OutputFormat              string        `long:"output-format" description:"default output when printint responses"`
+	SignModeStr               string        `long:"sign-mode" description:"sign mode to use"`
```
lesterli, line 23: These two fields' tag `mapstructure` should be changed to `long:xxx`

finality-provider/config/cosmwasm.go
```diff
@@ -7,20 +7,22 @@ import (
...
+	BabylonContractAddress    string        `mapstructure:"babylon-contract-address"`
+	BtcStakingContractAddress string        `mapstructure:"btc-staking-contract-address"`
 }
```
lesterli, line 26: These two fields' tag `mapstructure` should be changed to `long:xxx`

cosmwasmclient/config/config.go
```diff
@@ -7,40 +7,61 @@ import (
...
+	GasAdjustment             float64       `mapstructure:"gas-adjustment"`
+	GasPrices                 string        `mapstructure:"gas-prices"`
+	KeyDirectory              string        `mapstructure:"key-directory"`
+	Debug                     bool          `mapstructure:"debug"`
+	Timeout                   time.Duration `mapstructure:"timeout"`
+	BlockTimeout              time.Duration `mapstructure:"block-timeout"`
+	OutputFormat              string        `mapstructure:"output-format"`
+	SignModeStr               string        `mapstructure:"sign-mode"`
+	SubmitterAddress          string        `mapstructure:"submitter-address"`
+	BabylonContractAddress    string        `mapstructure:"babylon-contract-address"`
```
lesterli, line 32: Per Runchao's comment https://github.com/babylonchain/finality-provider/pull/379#discussion_r1643772395, the OPStack L2 consumer will also use the `Client` under the `cosmwasmclient`, so we should remove these two fields, and only keep them in the special config file of the cosmwasm consumer (now it is `finality-provider/config/cosmwasm.go`).

finality-provider/config/cosmwasm.go
gusin13: sure fixed

types/cosmwasm.go
gusin13: i see, we already have a types/config package outside the clientcontroller, if we have a similar thing inside wouldn't that be bit confusing?

lesterli: no, we can only create a separate package for each consumer, it would be like this:

SebastianElvis, line 1: Following the approach of https://github.com/babylonchain/finality-provider/pull/379, maybe the definitions could be moved to `clientcontroller/`?

clientcontroller/cosmwasm_consumer.go
```diff
@@ -4,20 +4,25 @@ import (
...
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 
 	sdkErr "cosmossdk.io/errors"
 	wasmdparams "github.com/CosmWasm/wasmd/app/params"
 	wasmdtypes "github.com/CosmWasm/wasmd/x/wasm/types"
+	bbn "github.com/babylonchain/babylon/types"
+	bbntypes "github.com/babylonchain/babylon/types"
```
SebastianElvis, line 17: The two dependencies are same

cosmwasmclient/config/config.go
SebastianElvis: Yep let's keep the `cosmwasmclient/` generic for all use cases. We can always feed the two contract addresses when submitting txs to the node

finality-provider/config/cosmwasm.go
```diff
@@ -7,55 +7,61 @@ import (
...
+		Timeout:                   20 * time.Second,
+		BlockTimeout:              1 * time.Minute,
+		OutputFormat:              "direct",
+		SignModeStr:               "",
+		BtcStakingContractAddress: "",
+		BabylonContractAddress:    "",
 	}
 }
 
 func ToQueryClientConfig(wc *CosmwasmConfig) *config.CosmwasmConfig {
```
SebastianElvis, line 48:
```suggestion
func (wc *CosmwasmConfig) ToQueryClientConfig() *config.CosmwasmConfig {
```

clientcontroller/cosmwasm_consumer.go
```diff
@@ -109,97 +115,321 @@ func (wc *CosmwasmConsumerController) CommitPubRandList(
...
+) (*fptypes.TxResponse, error) {
+	cmtProof := cmtcrypto.Proof{}
+	if err := cmtProof.Unmarshal(proof); err != nil {
+		return nil, err
+	}
+
+	var aunts []string
+	for _, aunt := range cmtProof.Aunts {
+		aunts = append(aunts, base64.StdEncoding.EncodeToString(aunt))
+	}
```
SebastianElvis, line 159: The base64 thing is really a mess :sweat: let's have a function that converts `cmtcrypto.Proof` to `fptypes.Proof`

SebastianElvis (commented):
The code lgtm! Given the refactoring in https://github.com/babylonchain/finality-provider/pull/379, the structure in this PR also needs to be updated. Let's wait until that PR is merged, then rebase and merge this PR. Wdyt?

clientcontroller/cosmwasm_consumer.go
maurolacy: Agreed. I'll see if we can have that removed for good.

itest/consumer_test_manager.go
```diff
@@ -70,6 +71,10 @@ func StartConsumerManager(t *testing.T) *ConsumerTestManager {
 	require.NoError(t, err)
 	cfg.CosmwasmConfig = config.DefaultCosmwasmConfig()
 	cfg.CosmwasmConfig.KeyDirectory = wh.dataDir
+	// make random contract addresses for now to avoid validation errors
```
maurolacy, line 74: Nit: Turn this into a TODO.

types/cosmwasm.go
```diff
@@ -0,0 +1,178 @@
...
+	SlashingTx           []byte                      `json:"slashing_tx"`
+	DelegatorSlashingSig []byte                      `json:"delegator_slashing_sig"`
+	CovenantSigs         []CovenantAdaptorSignatures `json:"covenant_sigs"`
+	StakingOutputIdx     uint32                      `json:"staking_output_idx"`
+	UnbondingTime        uint32                      `json:"unbonding_time"`
+	UndelegationInfo     *BtcUndelegationInfo        `json:"undelegation_info"`
+	ParamsVersion        uint32                      `json:"params_version"`
+}
+
+type SingleConsumerFpPowerResponse struct {
```
maurolacy, line 39: Nit: Rename this to `ConsumerFpResponse`.

types/cosmwasm.go
```diff
@@ -0,0 +1,178 @@
...
+	SubmitFinalitySignature SubmitFinalitySignature `json:"submit_finality_signature"`
+}
+
+type BtcStakingExecMsg struct {
+	BtcStaking BtcStaking `json:"btc_staking"`
+}
+
+type PubRandomnessExecMsg struct {
+	CommitPublicRandomness CommitPublicRandomness `json:"commit_public_randomness"`
+}
```
maurolacy, line 178: These could be part of a single `BtcStakingExecMsg`, with optional (pointer) fields for all the messages, and `omitempty` in the json description.
But I think it's the same / equivalent approach.

maurolacy (commented):
Good work with the queries!

types/cosmwasm.go
gusin13: but finality sig, pub randomness are not part of btc staking execute call?
https://github.com/babylonchain/babylon-contract/blob/474ad12b6f04a77b9c7f9a4567e50c528570a8e7/contracts/btc-staking/src/contract.rs#L160-L204

gusin13:
thanks for the review guys, i have made the suggested changes. Please have another look

clientcontroller/cosmwasm/consumer.go
```diff
@@ -56,45 +60,46 @@ func NewCosmwasmConsumerController(
...
 	addr, err := keyRec.GetAddress()
-
 	if err != nil {
-		panic(fmt.Sprintf("Failed to get key address: %s", err))
+		return nil, fmt.Errorf("failed to get address from key: %w", err)
 	}
 
-	return addr
+	return addr, nil
 }
```
lesterli, line 95: These two functions are unused, please remove them.

clientcontroller/cosmwasm/consumer.go
```diff
@@ -110,100 +115,304 @@ func (wc *CosmwasmConsumerController) CommitPubRandList(
...
+
+	return &fptypes.TxResponse{TxHash: res.TxHash}, nil
 }
 
 // QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height
 func (wc *CosmwasmConsumerController) QueryFinalityProviderVotingPower(fpPk *btcec.PublicKey, blockHeight uint64) (uint64, error) {
-	// empty response
-	return 0, nil
+	fpBtcPkHex := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex()
+	queryMsg := fmt.Sprintf(`{"finality_provider_info":{"btc_pk_hex":"%s", "height": %d}}`, fpBtcPkHex, blockHeight)
```
lesterli, line 232: It is better to have a `struct QueryMsg` to handle all msgs in CosmWasm, like:
```Go
type QueryMsg struct {
	FinalityProviderInfo *FinalityProviderInfo `json:"finality_provider_info,omitempty"`
}
```

clientcontroller/cosmwasm/consumer.go
```diff
@@ -110,100 +115,304 @@ func (wc *CosmwasmConsumerController) CommitPubRandList(
...
+
+	return blocks, nil
 }
 
-func (wc *CosmwasmConsumerController) QueryBlock(height uint64) (*types.BlockInfo, error) {
-	// TODO: dummy response, fetch actual block from the smart contract
-	block, err := wc.queryCometBestBlock()
+func (wc *CosmwasmConsumerController) queryIndexedBlock(height uint64) (*IndexedBlock, error) {
+	// Construct the query message
+	queryMsg := fmt.Sprintf(`{"block":{"height":%d}}`, height)
```
lesterli, line 325: same here

clientcontroller/cosmwasm/consumer.go
```diff
@@ -110,100 +115,304 @@ func (wc *CosmwasmConsumerController) CommitPubRandList(
...
+
+	// Return the finalized status
+	return resp.Finalized, nil
 }
 
 func (wc *CosmwasmConsumerController) QueryActivatedHeight() (uint64, error) {
-	// TODO: dummy response, fetch actual activated height from the smart contract
-	return 1, nil
+	// Construct the query message
+	queryMsg := `{"activated_height":{}}`
```
lesterli, line 375: same here

clientcontroller/cosmwasm/msg.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+type FinalitySigExecMsg struct {
+	SubmitFinalitySignature SubmitFinalitySignature `json:"submit_finality_signature"`
+}
+
+type BtcStakingExecMsg struct {
+	BtcStaking BtcStaking `json:"btc_staking"`
+}
+
```
lesterli, line 175:
```Go
type ExecMsg struct {
	SubmitFinalitySignature SubmitFinalitySignature `json:"submit_finality_signature"`
	BtcStaking BtcStaking `json:"btc_staking"`
	CommitPublicRandomness CommitPublicRandomness `json:"commit_public_randomness"`
}
```

clientcontroller/cosmwasm/consumer.go
```diff
@@ -110,100 +115,304 @@ func (wc *CosmwasmConsumerController) CommitPubRandList(
...
+	msg := FinalitySigExecMsg{
+		SubmitFinalitySignature: SubmitFinalitySignature{
+			FpPubkeyHex: bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex(),
+			Height:      block.Height,
+			PubRand:     base64.StdEncoding.EncodeToString(bbntypes.NewSchnorrPubRandFromFieldVal(pubRand).MustMarshal()),
+			Proof:       convertProof(cmtProof),
+			BlockHash:   base64.StdEncoding.EncodeToString(block.Hash),
+			Signature:   base64.StdEncoding.EncodeToString(bbntypes.NewSchnorrEOTSSigFromModNScalar(sig).MustMarshal()),
+		},
+	}
```
SebastianElvis, line 165: How about having a function `NewFinalitySigExecMsg` for constructing this message? Saw this piece of code a couple of times

finality-provider/config/cosmwasm.go
```diff
@@ -1,61 +1,100 @@
...
+		return fmt.Errorf("babylon-contract-address is required")
+	}
+	_, err := sdk.AccAddressFromBech32(cfg.BabylonContractAddress)
+	if err != nil {
+		return fmt.Errorf("babylon-contract-address: invalid bech32 address: %w", err)
+	}
+
+	if cfg.BtcStakingContractAddress == "" {
+		return fmt.Errorf("btc-staking-contract-address is required")
+	}
```
SebastianElvis, line 54: `AccAddressFromBech32` does this check already so we could remove it. Same for `BabylonContractAddress`

itest/test_helpers.go
```diff
@@ -10,133 +10,97 @@ import (
...
-type CovenantAdaptorSignatures struct {
-	CovPK       string   `json:"cov_pk"`
-	AdaptorSigs []string `json:"adaptor_sigs"`
-}
-
-type SignatureInfo struct {
-	PK  string `json:"pk"`
-	Sig string `json:"sig"`
-}
+func GenBtcStakingExecMsg(fpHex string) cosmwasm.BtcStakingExecMsg {
```
SebastianElvis, line 24: At this point, e2e test now has a number of setups (Babylon, OP stack, and CosmWasm). Would be good to have some separation bewteen those setups. For now, maybe we can move the datagen functions for CosmWasm integration to another file/folder?

itest/consumer_e2e_test.go
```diff
@@ -125,15 +115,15 @@ func TestSubmitFinalitySignature(t *testing.T) {
...
 	require.NotNil(t, dataFromContract)
-	var fpPower e2etypes.ConsumerFpsByPowerResponse
+	var fpPower cosmwasm.ConsumerFpsByPowerResponse
 	err = json.Unmarshal(dataFromContract.Data, &fpPower)
 	require.NoError(t, err)
 	require.Len(t, fpPower.Fps, 1)
 	require.Equal(t, msg.BtcStaking.NewFP[0].BTCPKHex, fpPower.Fps[0].BtcPkHex)
 	require.Equal(t, consumerDels.ConsumerDelegations[0].TotalSat, fpPower.Fps[0].Power)
 
 	// inject pub rand commitment in smart contract (admin is not required, although in the tests admin and sender are the same)
```
SebastianElvis, line 125: In theory FP should inject pub rand commitment automatically and we don't need to do this manually. Is the reason that we need https://github.com/babylonchain/finality-provider/pull/382 first?

SebastianElvis (approved):
Great work!

itest/consumer_e2e_test.go
gusin13: this is manually done because finality provider daemon is not working in the e2e tests. 382 is just completing the randomness query

SebastianElvis: I see. Do you think we need a e2e test where fpd is running and executing these automatically?

maurolacy (approved):
LGTM.

types/cosmwasm.go
maurolacy: They aren't. But all of them are part of the `ExecuteMsg` itself. Anyway, it's the same result in the end. No need to change this.

clientcontroller/cosmwasm/msg.go
maurolacy: Yes, this is what I was talking about. I think those have to be pointers, by the way.

Both approaches are in the equivalent, and in mesh-security by example, both are being used.

itest/consumer_e2e_test.go
gusin13: https://github.com/babylonchain/finality-provider/issues/391

itest/test_helpers.go
gusin13: the datagen things are already in test_helpers, maybe when i setup bcd i need to restructure files will figure out a better place for this then.

clientcontroller/cosmwasm/consumer.go
gusin13: i have refactored this as per 
https://github.com/babylonchain/finality-provider/pull/381#discussion_r1646735762

PR #382, opened 2024-06-18
chore: bump babylon and fix pub rand consumer query

gusin13:
Follow up pr after https://github.com/babylonchain/finality-provider/pull/381

- incorporates pub randomness cosmwasm consumer query
- bumps babylon as the response type of pub randomness grpc query has changed in babylon


types/pub_rand_commit.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/cometbft/cometbft/crypto/merkle"
 )
 
+type PubRandCommit struct {
+	NumPubRand uint64 `json:"num_pub_rand"`
+	Commitment []byte `json:"commitment"`
+}
```
SebastianElvis, line 12: Why is start height omitted? It seems to be necessary for positioning a pub rand commit

SebastianElvis (approved):
Code lgtm :+1: would be good if we could rebase this PR and https://github.com/babylonchain/finality-provider/pull/381 before merging

maurolacy (approved):
LGTM.

types/pub_rand_commit.go
gusin13: yep correct as Mauro mentioned this is a common output struct shared by all consumers (babylon, evm, cosmwasm), when the pub randomness is retrieved from the underlying consumer it is then converted to a map of height and this struct
`map[uint64]*types.PubRandCommit`
we could have the height in this struct but that is redundant as the height is key of map

PR #384, opened 2024-06-19
fix: typo in OPStack clientcontroller

gusin13:
Minor typo in client

### DIFF
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -32,7 +32,7 @@ const (
 var _ api.ConsumerController = &OPStackL2ConsumerController{}
 
 type OPStackL2ConsumerController struct {
-	cwcClient  *cwclient.Client
+	cwClient   *cwclient.Client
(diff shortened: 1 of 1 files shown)


clientcontroller/opstackl2/consumer.go
```diff
@@ -58,7 +58,7 @@ func NewOPStackL2ConsumerController(
 	bbnClient, err := cwclient.New(
 		&cwConfig,
 		BabylonChainName,
```
gusin13, line 60: another thing i noticed we have this constant
`BabylonChainName` in this file and then another one 
`BabylonConsumerChainName` in clientcontroller/factory
are those same?

lesterli (approved):
Thanks

clientcontroller/opstackl2/consumer.go
lesterli: no, I think they represent different meanings so I didn't just use the existing variable `BabylonConsumerChainName`.

PR #394, opened 2024-06-21
chore: fix nil pointer

SebastianElvis:
When error occurs, `block` is `nil` and `block.Height` will cause a nil pointer panic. Uncovered in local deployment


maurolacy (approved):
LGTM.

maurolacy:
You might want to backport this fix to the `euphrates-0.2.x` branch as well.

SebastianElvis:
> You might want to backport this fix to the `euphrates-0.2.x` branch as well.

This does not exist in `euphrates-0.2.x`, so no need to

PR #401, opened 2024-06-22
fix: fast sync for consumer

lesterli:
This PR fixes the error of fast sync for the consumer.
When the finality provider is not activated yet on the consumer, there is no need to catch up.
Also, we checked all where the function `tryFastSync` is called, and it would not block/break them when returning `nil`.

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -363,6 +363,20 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlockHeight uint64) (*Fast
 		return nil, fmt.Errorf("the finality-provider %s is already in sync", fp.GetBtcPkHex())
 	}
 
+	// get the activated height from the consumer controller
+	activatedHeight, err := fp.consumerCon.QueryActivatedHeight()
+	if err != nil {
+		return nil, err
+	}
+	if targetBlockHeight < activatedHeight {
+		fp.logger.Debug(
+			"finality gadget is not activated yet, no need to catch up",
+			zap.Uint64("activated height", activatedHeight),
+			zap.Uint64("target height", targetBlockHeight),
+		)
+		return nil, nil
+	}
+
 	// get the last finalized height
 	lastFinalizedBlock, err := fp.latestFinalizedBlockWithRetry()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bap2pecs (approved):

bap2pecs:
@SebastianElvis this was not an issue b/c we don't have a use case yet when the consumer chain has been running for a long time. so we always start from block 0 in fast sync.

but imagine if we want to add our system to a chain that has been running for years, it's unnecessary to start from block 0

finality-provider/service/fp_instance.go
```diff
@@ -363,6 +363,20 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlockHeight uint64) (*Fast
 		return nil, fmt.Errorf("the finality-provider %s is already in sync", fp.GetBtcPkHex())
 	}
 
+	// get the activated height from the consumer controller
+	activatedHeight, err := fp.consumerCon.QueryActivatedHeight()
+	if err != nil {
+		return nil, err
+	}
+	if targetBlockHeight < activatedHeight {
```
SebastianElvis, line 371: I think there is a confusion for the activation of BTC staking. We say BTC staking is activated when Babylon receives the 1st active BTC delegation. It's not relevant to whether a FP has committed a pub rand yet. 

Here, a proper fix imo would be:

1. In DA contract, if no pub rand commit is found, it returns nil without any error. Currently it returns error which causes the problem. If you look at FP code, FP actually handles the case when it never commits pub rand.
2. In this file, if FP finds the error being sth like "no pub rand commit is found", we don't return error here.

Suggest we do 1. Wdyt?

lesterli: yes, this PR fixes it. https://github.com/babylonchain/babylon-contract/pull/184

bap2pecs: > I think there is a confusion for the activation of BTC staking. We say BTC staking is activated when Babylon receives the 1st active BTC delegation. It's not relevant to whether a FP has committed a pub rand yet. 
> 
> 
> 
> Here, a proper fix imo would be:
> 
> 
> 
> 1. In DA contract, if no pub rand commit is found, it returns nil without any error. Currently it returns error which causes the problem. If you look at FP code, FP actually handles the case when it never commits pub rand.
> 
> 2. In this file, if FP finds the error being sth like "no pub rand commit is found", we don't return error here.
> 
> 
> 
> Suggest we do 1. Wdyt?

yes we realized and reverted this PR. 1 is also implemented

Issue #397, opened 2024-06-21
When e2e failed, background daemons are still running


bap2pecs:
this gave us lots of trouble when debugging a failed test

```
$ make test-e2e-op                                                                             [18:13:35]
cd tools; \
        go install -trimpath github.com/babylonchain/babylon/cmd/babylond
go test -mod=readonly -timeout=25m -v github.com/babylonchain/finality-provider/itest github.com/babylonchain/finality-provider/itest/opstackl2 -count=1 --tags=e2e_op
?       github.com/babylonchain/finality-provider/itest [no test files]
=== RUN   TestSubmitFinalitySignature
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
    test_manager.go:105: Babylon node is started
2024/06/21 18:17:56 Cannot remove dir 1
2024/06/21 18:17:56 Cannot remove dir 2
    test_manager.go:110: 
                Error Trace:    /Users/<redacted>/Documents/Projects/babylon-finality-provider/itest/opstackl2/test_manager.go:110
                                                        /opt/homebrew/Cellar/go/1.22.4/libexec/src/runtime/panic.go:770
                                                        /Users/<redacted>/Documents/Projects/babylon-finality-provider/cosmwasmclient/client/keys.go:18
                                                        /Users/<redacted>/Documents/Projects/babylon-finality-provider/itest/opstackl2/e2e_test.go:42
                                                        /Users/<redacted>/Documents/Projects/babylon-finality-provider/itest/opstackl2/e2e_test.go:77
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestSubmitFinalitySignature
--- FAIL: TestSubmitFinalitySignature (1.88s)
FAIL
FAIL    github.com/babylonchain/finality-provider/itest/opstackl2       2.833s
FAIL
make: *** [test-e2e-op] Error 1
```

then we realized it's b/c there were some process running:

```
$ ps                                                                                           [18:17:56]
  PID TTY           TIME CMD
 8321 ttys001    2:12.87 babylond start --home=/var/folders/9_/q4wsdnh14_s60_74cd2rbztm0000gp/T/zBabylonTest2191261572/node0/babyl
[...]

PR #405, opened 2024-06-22
chore: add `clean-e2e` command to kill dangling babylond or wasmd processed when e2e tests fails

bap2pecs:
context https://github.com/babylonchain/finality-provider/issues/397

note that this just a workaround, we probably should avoid this happen in the first place

but this PR is good for now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,11 +67,17 @@ install-wasmd:
 	cd $(TOOLS_DIR); \
 	go install -trimpath $(WASMD_PKG)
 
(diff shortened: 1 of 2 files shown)


lesterli (approved):

SebastianElvis:
Nice fix! Should we cherry-pick this to `dev` as well? @gitferry

bap2pecs:
additional context, this command cannot run below the go test commands b/c it will break the running tests (i tried)

it has to put in the beginning

PR #406, opened 2024-06-22
refactor: expose a few private names from e2e test

bap2pecs:
## Summary

our own e2e test will use those names

move them to utils.go

in the future, we should refactor to create independent packages for opl2, wasm, and babylon consumers but that's outside of the scope of this PR

## Test Plan

```
make test-e2e
make test
```

### DIFF
--- a/itest/test_manager.go
+++ b/itest/test_manager.go
@@ -25,7 +25,6 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
-	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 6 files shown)


lesterli (approved):

itest/test_manager.go
```diff
@@ -38,16 +37,8 @@ import (
...
-	btcNetworkParams      = &chaincfg.SimNetParams
-
-	fpNamePrefix  = "test-fp-"
-	monikerPrefix = "moniker-"
-	chainID       = "chain-test"
-	passphrase    = "testpass"
-	hdPath        = ""
-	simnetParams  = &chaincfg.SimNetParams
+	btcNetworkParams = &chaincfg.SimNetParams
+	simnetParams     = &chaincfg.SimNetParams
```
SebastianElvis, line 41: Should they be merged?

lesterli: Yes, fixed https://github.com/babylonchain/finality-provider/pull/409

PR #408, opened 2024-06-23
fix: parse cw data with null

lesterli:
## Summary

This PR fixes the issue of parsing cw data with null. When a cw contract's return data is `None`, the JSON is a four-character string `null` and the `json.Unmarshal()` does NOT raise an error. 

## Test Plan

make test-e2e


bap2pecs (approved):

bap2pecs:
unfortuantely unmarshall of "null" will result in 

```
"0": {
    "num_pub_rand": 0,
    "commitment": null
}
```

Issue #415, opened 2024-06-23
missing e2e tag in many itest files


bap2pecs:
```
//go:build e2e
// +build e2e
```

is missing in many files inside `itest/`

for example, `itest/consumer_e2e_test.go` has the tag but `itest/consumer_test_manager.go` doesn't have it

this can confuse lint

https://github.com/babylonchain/finality-provider/issues/414 is related and I suggest to fix the two issues together

SebastianElvis:
Only files that contain test cases need the `e2e` tag while stuff like `itest/consumer_test_manager.go` does not need it, no?

bap2pecs:
@SebastianElvis not exactly. we also followed the current structure initially to just add the tag in `itest/opstackl2/op_e2e_test.go` but not in `itest/opstackl2/op_test_manager.go`

however, CI failed complaining about these two vars not used:

```
const (
	opFinalityGadgetContractPath = "../bytecode/op_finality_gadget_f149c8b.wasm"
	opConsumerId                 = "op-stack-l2-12345"
)
```

digging deeper, it's b/c in the Makefil `PACKAGES_E2E=$(shell go list ./... | grep '/itest')` will exclude the files with tags

so when linting, it won't see the test file and see those two vars not used.

I think this implies two things:

- we should tag both files in the `e2etest_op` package with the same tag
- currently CI linting fails to cover certain files

so I did two things to fix it:
- add the tag in `op_test_manager.go`
- fixed the Makefile to add and use `PACKAGES_E2E_OP=$(shell go list -tags=e2e_op ./... | grep '/itest')`

cc @maurolacy this also explained you other question in https://github.com/babylonchain/finality-provider/pull/418#pullrequestreview-2134979393

gusin13:
hmm i see, ideally shouldn't we also lint the files which have build tags? I see there are 2 problems 

1. We want the test files to be excluded from the go build i.e why we use // go:build <some_tag> so the binary doesn't include these files
2. The golangci-lint excludes files with go build 

Solution 
1. Either we should add the go build tag to all the test related files (including test managers etc or any helpers) this is enhancement of what @bap2pecs mentioned above
2. Or just add the build tag to _test files and also enable the linter to include files with build tag

wdyt?

bap2pecs:
> ideally shouldn't we also lint the files which have build tags

[...]

PR #418, opened 2024-06-23
test: start FP in op e2e

bap2pecs:
## Summary

start FP in e2e test

also found two problems"

1. incorrect wasm binary was used. it might be overwritten when I was fixing merge conflicts. a better way is to append the binary with the commit hash where I built from in https://github.com/babylonchain/babylon-contract
2. `go list` can't find any test b/c tag was not added

## Test Plan

```
make test-e2e-op
```

wait for CI


lesterli (approved):

bap2pecs:
@gusin13 I noticed that there is a version.txt file you added in https://github.com/babylonchain/finality-provider/pull/374

any tip how we should update it given that we also have our own bytecode added?

Makefile
```diff
@@ -83,11 +84,11 @@ clean-e2e:
 test-e2e: install-babylond install-wasmd
 	make clean-e2e
 	go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E) -count=1 --tags=e2e
-	go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E) -count=1 --tags=e2e_op
+	go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op
```
maurolacy, line 87: Is this correct? Looks like a bug to me.

bap2pecs: which part looks wrong? I added some explanation in https://github.com/babylonchain/finality-provider/issues/415

maurolacy: Shouldn't the env var be `PACKAGES_E2E` here? I mean, these tests are the e2e tests, not the e2e-op ones.

Perhaps I'm missing something.

bap2pecs: `test-e2e` is supposed to test all tests including e2e-op ones

maurolacy: Got it, there's another `go test` above.

PR #423, opened 2024-06-24
chore: refactor test directories

gusin13:
Refactors dirs

Screenshot 2024-06-24 at 5 32 02 PM
Screenshot 2024-06-24 at 5 32 16 PM


lesterli:
@gusin13 Can we keep all common files under the `itest` and use the package `e2etest`? So that others can still import them with the `e2etest.`.

SebastianElvis:
> @gusin13 Can we keep all common files under the `itest` and use the package `e2etest`? b/c no need changes for the callers which import them with the `e2etest`

Another way to structure the e2e test is that each package includes all utility functions, e.g., node handlers, while `itest/` contains all test cases. This way it's more clear which tests does `itest/` contain, and there won't be potential cyclic dependency issues where two packages depend on each other (think about a scenario where wasm e2e uses babylon while babylon uses wasm). 

But the current structure also looks good to me :+1:

SebastianElvis (approved):
Great work! The PR needs rebasing though

itest/common/utils.go
```diff
@@ -0,0 +1,95 @@
+package e2etest_common
```
maurolacy, line 1: 👍🏼 `e2etest_utils` is also a good name here.

maurolacy (approved):
LGTM, nice refactor.

maurolacy:
Just seen that there are some conflicts with the base branch. That could take a while.

bap2pecs (approved):
LGTM as well! only suggestion is we can consider to do sth like this

```
PACKAGES_E2E_OP=$(shell go list -tags=e2e_op ./... | grep '/itest')
```

the current var
```
PACKAGES_E2E=$(shell go list ./... | grep '/itest')
```
isn't technically correct. it worked only because there are files that don't have the build tags in /itest

but if we fix https://github.com/babylonchain/finality-provider/issues/415 using approach 1, the Makefile will need to be fixed as well

gusin13:
Thanks for the review guys. 

@lesterli I have transferred files from itest/common to itest/ now 

@SebastianElvis for now i have kept the same structure, but agree with your suggestion as well. If there are cyclic deps in future we can transfer the babylon node handler and other common things under each package.

@bap2pecs i was testing few things about this issue 415, had a doubt if i do
```
test-e2e-op: clean-e2e install-babylond
	go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E) -count=1 --tags=e2e_op
```

this still works, i thought it would not work. So if thats the case why do we even need `PACKAGES_E2E_OP`

PR #426, opened 2024-06-25
fix: op e2e test

lesterli:
## Summary

This PR fixes the following issues:

* `ErrNoBabylonFPRestaked`, which is caused by no Babylon FP, a [BTC delegation](https://github.com/babylonchain/babylon-private/blob/base/consumer-chain-support/x/btcstaking/keeper/msg_server.go#L169-L213
) has to stake to at least one Babylon finality provider.
* `TooFewPubRand`, which is caused by a race condition, when FP starts, it will start the pub rand submission loop. However, we also manually inject the pub rands into the CW contract in our test case. It would throw the `TooFewPubRand` error, such as: 
   ```
   The start height (1) has overlap with the height of the highest public randomness committed (<a height submitted by the FP submission loop>)
   ```

## Test Plan

```
make lint
make test-e2e
```


itest/opstackl2/op_e2e_test.go
```diff
@@ -18,32 +21,53 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	ctm := StartOpL2ConsumerManager(t)
 	defer ctm.Stop(t)
 
-	// register FP in Babylon Chain
-	fpList := ctm.StartFinalityProvider(t, 1)
+	// start Babylon chain FP
```
bap2pecs, line 24: let's add some context b/c it's not v easy to understand why we need it

itest/opstackl2/op_e2e_test.go
```diff
@@ -18,32 +21,53 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
+	_ = ctm.StartFinalityProvider(t, e2etest.ChainID, 1)
```
bap2pecs, line 25: why do we need to use `e2etest.ChainID`? does any logic depend on this specific value? if not, we should just use a mock value

bap2pecs: also let's remove `_=`

itest/opstackl2/op_e2e_test.go
```diff
@@ -18,32 +21,53 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
+	fpList := ctm.StartFinalityProvider(t, opConsumerId, 1)
+	var fpInstance *service.FinalityProviderInstance
+	for _, fp := range fpList {
+		if bytes.Equal(fp.GetChainID(), []byte(opConsumerId)) {
+			fpInstance = fp
+			break
+		}
+	}
+
+	ctm.WaitForFpPubRandCommitted(t, fpInstance)
```
bap2pecs, line 37: let's add a new line after this line b/c `WaitForFpPubRandCommitted` is separate from the next section "query pub rand"

itest/opstackl2/op_e2e_test.go
```diff
@@ -18,32 +21,53 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
+		lastCommittedStartHeight = key
+		break
+	}
+	t.Logf("Last committed pubrandList startHeight %d", lastCommittedStartHeight)
+	pubRandList, err := fpInstance.GetPubRandList(lastCommittedStartHeight, ctm.FpConfig.NumPubRand)
+	require.NoError(t, err)
+	// generate commitment and proof for each public randomness
+	_, proofList := types.GetPubRandCommitAndProofs(pubRandList)
 
 	// mock block
```
bap2pecs, line 52: mock block hash

itest/opstackl2/op_e2e_test.go
```diff
@@ -18,32 +21,53 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
-	ctm.WaitForFpPubRandCommitted(t, fpList[0].GetBtcPkBIP340())
+	// start consumer chain FP
+	fpList := ctm.StartFinalityProvider(t, opConsumerId, 1)
+	var fpInstance *service.FinalityProviderInstance
+	for _, fp := range fpList {
+		if bytes.Equal(fp.GetChainID(), []byte(opConsumerId)) {
+			fpInstance = fp
+			break
+		}
+	}
```
bap2pecs, line 35: let's avoid this for loop since we already know there is only one FP from line 28

Edit: discussed offline. the confusion was `StartFinalityProvider` doesn't return the right number of FPs. we should move this for loop inside and enforce the returned list length is `n`

itest/opstackl2/op_test_manager.go
```diff
@@ -216,154 +215,41 @@ func (ctm *OpL2ConsumerTestManager) StartFinalityProvider(t *testing.T, n int) [
...
 				}
+				return len(fps) > 0
 			}
-
-			return true
 		}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
 	}
 
 	fpInsList := app.ListFinalityProviderInstances()
-	require.Equal(t, n, len(fpInsList))
```
bap2pecs, line 245: let's add this back per our discussion

itest/opstackl2/op_test_manager.go
```diff
@@ -216,154 +215,41 @@ func (ctm *OpL2ConsumerTestManager) StartFinalityProvider(t *testing.T, n int) [
...
-
 	t.Logf("The test manager is running with %v finality-provider(s)", len(fpInsList))
 
-	for i, fp := range fpInsList {
```
bap2pecs, line 249: we can also add this back

itest/opstackl2/op_test_manager.go
```diff
@@ -216,154 +215,41 @@ func (ctm *OpL2ConsumerTestManager) StartFinalityProvider(t *testing.T, n int) [
...
 		require.Eventually(t, func() bool {
-			fps, err := ctm.QueryFinalityProviders(opConsumerId)
-			if err != nil {
-				t.Logf("failed to query finality providers from Babylon %s", err.Error())
-				return false
-			}
-
-			if len(fps) != i+1 {
-				return false
-			}
```
bap2pecs, line 229: seems now we can also move back this and `QueryFinalityProviders`?

itest/opstackl2/op_e2e_test.go
lesterli: > why do we need to use `e2etest.ChainID`? does any logic depend on this specific value? if not, we should just use a mock value

Yes, we need to use it, while using a mock value, it throws the error:
```
Error:          Received unexpected error:
                the finality-provider manager has already stopped
```

bap2pecs:
let's also add more context in the description e.g. why the race condition didn't happen

bap2pecs (approved):

itest/opstackl2/op_test_manager.go
```diff
@@ -191,18 +190,24 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
 	t.Logf("Babylon node has started")
 }
 
-func (ctm *OpL2ConsumerTestManager) StartFinalityProvider(t *testing.T, n int) []*service.FinalityProviderInstance {
+func (ctm *OpL2ConsumerTestManager) StartFinalityProvider(t *testing.T, isBabylonFp bool, n int) []*service.FinalityProviderInstance {
 	app := ctm.FpApp
 
+	chainId := opConsumerId
+	if isBabylonFp {
+		// While using another mock value, it throws the error: the finality-provider manager has already stopped
```
bap2pecs, line 198: found the code where magic happens: https://github.com/babylonchain/babylon-private/blob/base/consumer-chain-support/x/btcstaking/keeper/msg_server.go#L100-L106

```
	if consumerID == ctx.ChainID() {
		ms.SetFinalityProvider(ctx, &fp)
	} else {
		if err := ms.SetConsumerFinalityProvider(ctx, &fp, consumerID); err != nil {
			return nil, err
		}
	}
```

bap2pecs:
add some more context, we didnt get any race condition earlier b/c the default PR submission interval is 30 seconds

so we changed it to `cfg.RandomnessCommitInterval = 500 * time.Millisecond` and saw the problem

but this PR fixed it

PR #429, opened 2024-06-25
test: handle multiple providers in `InsertCovenantSigForDelegation`

bap2pecs:
## Summary

we need to use it in our own e2e to pass in both a Babylon FP and a Consumer FP

## Test Plan

```
make test-e2e-babylon   
```


lesterli (approved):
Lgtm

Issue #407, opened 2024-06-23
improve interface `QueryLastCommittedPublicRand()`


bap2pecs:
```
	QueryLastCommittedPublicRand(fpPk *btcec.PublicKey, count uint64) (map[uint64]*types.PubRandCommit, error)
```

it will always pass in count `1`

and the return value shouldn't be a map because it will never have two keys in the map

for now it will always return a single-value map like this:

```
{
   "122881": {"num_pub_rand": 64, "commitment": <hash>}
}
```

so I'd suggest we remove the `count` and simplify the return type

lesterli:
Indeed, we should consider refactoring if it will never have two keys on the map. b/c it is inconvenient to use, the codes would be like this:

```Go
	// query pub rand
	committedPubRandMap, err := ctm.OpL2ConsumerCtrl.QueryLastCommittedPublicRand(fpInstance.GetBtcPk(), 1)
	require.NoError(t, err)
	var lastCommittedHeight uint64
	for key, _ := range committedPubRandMap {
		lastCommittedHeight = key
		break
	}
```

SebastianElvis:
Sounds good to me :+1:

PR #436, opened 2024-06-26
refactor: clean up interface `QueryLastCommittedPublicRand()`

bap2pecs:
## Summary

#407 

## Test Plan

```
make lint
```
wait for CI

### DIFF
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -297,16 +297,21 @@ func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.Pu
 		return nil, fmt.Errorf("failed to query committed public randomness: %w", err)
 	}
 
-	var commit types.PubRandCommit
+	if len(res.PubRandCommitMap) != 1 {
+		// before first PR commitment, it's expected to return nothing
+		return nil, nil
+	}
+
+	var commit *types.PubRandCommit
 	for height, commitRes := range res.PubRandCommitMap {
-		commit = types.PubRandCommit{
+		commit = &types.PubRandCommit{
 			StartHeight: height,
 			NumPubRand:  commitRes.NumPubRand,
 			Commitment:  commitRes.Commitment,
 		}
 	}
 
-	return &commit, nil
+	return commit, nil
 }
 
 func (bc *BabylonConsumerController) QueryIsBlockFinalized(height uint64) (bool, error) {
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -298,7 +298,7 @@ func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.Pu
 	}
 
 	if len(res.PubRandCommitMap) > 1 {
-		return nil, fmt.Errorf("expected length to be 1, but got :%d", len(res.PubRandCommitMap))
+		return nil, fmt.Errorf("expected length to be 1, but get :%d", len(res.PubRandCommitMap))
 	}
 
 	var commit *types.PubRandCommit
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -297,11 +297,16 @@ func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.Pu
 		return nil, fmt.Errorf("failed to query committed public randomness: %w", err)
 	}
 
+	if len(res.PubRandCommitMap) == 0 {
+		// expected when there is no PR commit at all
+		return nil, nil
+	}
+
 	if len(res.PubRandCommitMap) > 1 {
 		return nil, fmt.Errorf("expected length to be 1, but get :%d", len(res.PubRandCommitMap))
 	}
 
-	var commit *types.PubRandCommit
+	var commit *types.PubRandCommit = nil
 	for height, commitRes := range res.PubRandCommitMap {
 		commit = &types.PubRandCommit{
 			StartHeight: height,
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -283,8 +283,8 @@ func (bc *BabylonConsumerController) QueryBlock(height uint64) (*types.BlockInfo
 	}, nil
 }
 
-// QueryLastCommittedPublicRand returns the last public randomness commitments
-func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.PublicKey) (*types.PubRandCommit, error) {
+// QueryLastPublicRandCommit returns the last public randomness commitments
+func (bc *BabylonConsumerController) QueryLastPublicRandCommit(fpPk *btcec.PublicKey) (*types.PubRandCommit, error) {
 	fpBtcPk := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
 
 	pagination := &sdkquery.PageRequest{
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -297,9 +297,8 @@ func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.Pu
 		return nil, fmt.Errorf("failed to query committed public randomness: %w", err)
 	}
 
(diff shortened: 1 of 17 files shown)


types/pub_rand_commit.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 type PubRandCommit struct {
-	NumPubRand uint64 `json:"num_pub_rand"`
-	Commitment []byte `json:"commitment"`
+	StartHeight uint64 `json:"start_height"`
```
SebastianElvis, line 10: Why is this added back (although I do prefer to have it)?

clientcontroller/babylon/consumer.go
```diff
@@ -298,15 +297,16 @@ func (bc *BabylonConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.Pu
 		return nil, fmt.Errorf("failed to query committed public randomness: %w", err)
 	}
 
-	commitMap := make(map[uint64]*types.PubRandCommit)
-	for height, commit := range res.PubRandCommitMap {
-		commitMap[height] = &types.PubRandCommit{
-			NumPubRand: commit.NumPubRand,
-			Commitment: commit.Commitment,
+	var commit types.PubRandCommit
+	for height, commitRes := range res.PubRandCommitMap {
```
SebastianElvis, line 301: Let's play safe here: if `len(res.PubRandCommitMap) == 1`, do the following, otherwise return `nil, nil`.

Currently `commit` might be given an empty (but not nil) value, not 100% sure. Maybe this is also why e2e fails?

clientcontroller/cosmwasm/consumer.go
```diff
@@ -421,16 +422,17 @@ func (wc *CosmwasmConsumerController) QueryLastCommittedPublicRand(fpPk *btcec.P
...
 
 	// Convert the response to the expected map format
-	commitMap := make(map[uint64]*fptypes.PubRandCommit)
-	for _, commit := range commits {
-		commitCopy := commit // create a copy to avoid referencing the loop variable
-		commitMap[commit.StartHeight] = &fptypes.PubRandCommit{
-			NumPubRand: commitCopy.NumPubRand,
-			Commitment: commitCopy.Commitment,
+	var commit *fptypes.PubRandCommit
+	for _, commitRes := range commits {
```
SebastianElvis, line 426: Same here

types/pub_rand_commit.go
bap2pecs: b/c now we don't need both
```
type PubRandCommit struct {
	NumPubRand uint64 `json:"num_pub_rand"`
	Commitment []byte `json:"commitment"`
}
```

and

```
type LastPubRandCommitResponse struct {
	StartHeight uint64 `json:"start_height"`
	NumPubRand  uint64 `json:"num_pub_rand"`
	Commitment  []byte `json:"commitment"`
}
```

clientcontroller/babylon/consumer.go
lesterli: Yes, indeed the reason is that `commit` is an empty value but not `nil`.

bap2pecs: this is probably a better condition:

```
	if len(res.PubRandCommitMap) > 1 {
		return nil, fmt.Errorf("expected length to be 1, but got :%d", len(res.PubRandCommitMap))
	}

	var commit *types.PubRandCommit
	for height, commitRes := range res.PubRandCommitMap {
		commit = &types.PubRandCommit{
			StartHeight: height,
			NumPubRand:  commitRes.NumPubRand,
			Commitment:  commitRes.Commitment,
		}
	}

	return commit, nil
```

SebastianElvis: Shouldn't we still check whether `len(res.PubRandCommitMap) == 0` and returns `nil, nil` if this is true?

bap2pecs: wait I think the above code will return nil nil when it's 0 length. is it not true?

SebastianElvis: Right, `commit` is now a pointer so in theory it will be `nil` if `res` is empty. Personally I feel it's a bit unintuitive so prefer `var commit *types.PubRandCommit = nil`, but no strong opinion, your call

SebastianElvis (approved):
Great work! Would be good to get feedback from @gitferry as well since this is an interface level change

finality-provider/service/fp_instance.go
```diff
@@ -891,31 +891,29 @@ func (fp *FinalityProviderInstance) getPollerStartingHeight() (uint64, error) {
...
 func (fp *FinalityProviderInstance) GetLastCommittedHeight() (uint64, error) {
-	pubRandCommitMap, err := fp.lastCommittedPublicRandWithRetry(1)
+	pubRandCommit, err := fp.lastCommittedPublicRandWithRetry()
 	if err != nil {
 		return 0, err
 	}
 
 	// no committed randomness yet
-	if len(pubRandCommitMap) == 0 {
+	if pubRandCommit == nil || (pubRandCommit.StartHeight == 0 && pubRandCommit.NumPubRand == 0) {
```
gitferry, line 900: This would be more readable if we have a helper method of `PubRandCommit` like `pubRandCommit.IsEmpty()`

gitferry (approved):
Great work! Only nits

finality-provider/service/fp_instance.go
bap2pecs: I am a bit hesitant to create helper function b/c when it's empty it implies a bug in the consumer interface implementation. see conv here https://github.com/babylonchain/finality-provider/pull/438#discussion_r1657077298

clientcontroller/babylon/consumer.go
bap2pecs: ok added both to make it more readable:

```
	if len(res.PubRandCommitMap) == 0 {
		// expected when there is no PR commit at all
		return nil, nil
	}

	if len(res.PubRandCommitMap) > 1 {
		return nil, fmt.Errorf("expected length to be 1, but get :%d", len(res.PubRandCommitMap))
	}

	var commit *types.PubRandCommit = nil
	for height, commitRes := range res.PubRandCommitMap {
		commit = &types.PubRandCommit{
			StartHeight: height,
			NumPubRand:  commitRes.NumPubRand,
			Commitment:  commitRes.Commitment,
		}
	}

	return commit, nil
```

clientcontroller/cosmwasm/consumer.go
bap2pecs: changed to 

```
	if len(commits) == 0 {
		// expected when there is no PR commit at all
		// `get_pub_rand_commit`'s return type is Vec<PubRandCommit> and it can be
		// empty vector if no results found
		return nil, nil
	}

	if len(commits) > 1 {
		return nil, fmt.Errorf("expected length to be 1, but got :%d", len(commits))
	}

	// Convert the response to the expected map format
	var commit *fptypes.PubRandCommit = nil
	for _, commitRes := range commits {
		commitCopy := commitRes // create a copy to avoid referencing the loop variable
		commit = &fptypes.PubRandCommit{
			StartHeight: commitCopy.StartHeight,
			NumPubRand:  commitCopy.NumPubRand,
			Commitment:  commitCopy.Commitment,
		}
	}

	return commit, nil
```

cc @maurolacy

finality-provider/service/fp_instance.go
gitferry: If so, I think it makes more sense to move the check to the client controller implementation

bap2pecs: maybe we can have sth like this:

```
func (prc *PubRandCommit) IsValid() bool {
    return prc.NumPubRand >= 1
}
```

wdyt? @gitferry

gitferry: Close to my point replied in https://github.com/babylonchain/finality-provider/pull/438#discussion_r1657169696. Maybe `Validate() error` is more informative

bap2pecs: > If so, I think it makes more sense to move the check to the client controller implementation

the problem I discussed w @SebastianElvis was the client impl can be incorrect and we want to add some check to detect it here

but I do see we can move it to `lastCommittedPublicRandWithRetry` at least

gitferry: Sounds good to me

PR #437, opened 2024-06-26
fix: op e2e test

lesterli:
This PR fixes the op e2e test issue which was caused by the block hash with the wrong format.

## Test Plan

```
make test
make test-e2e
```

### DIFF
--- a/itest/opstackl2/op_e2e_test.go
+++ b/itest/opstackl2/op_e2e_test.go
@@ -25,7 +25,8 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	fpList := ctm.StartFinalityProvider(t, false, 1)
 	fpInstance := fpList[0]
 
-	e2eutils.WaitForFpPubRandCommitted(t, fpInstance)
+	// expect 2 rounds of submissions
+	e2eutils.WaitForFpPubRandCommitted(t, fpInstance, 2)
 
 	// query pub rand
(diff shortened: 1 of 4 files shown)


SebastianElvis (approved):

itest/opstackl2/op_e2e_test.go
```diff
@@ -139,49 +138,50 @@ func TestBlockBabylonFinalized(t *testing.T) {
...
 		t.Logf("Last committed pubrandList startHeight %d", lastCommittedStartHeight)
 
 		pubRandList, err := fp.GetPubRandList(lastCommittedStartHeight, ctm.FpConfig.NumPubRand)
 		require.NoError(t, err)
 		// generate commitment and proof for each public randomness
 		_, proofList := types.GetPubRandCommitAndProofs(pubRandList)
 
-		// mock block hash
-		r := rand.New(rand.NewSource(1))
-		mockHash := datagen.GenRandomByteArray(r, 32)
```
bap2pecs, line 157: this method is a bit dangerous to use jfyi @SebastianElvis @gitferry 

this PR is to actually fix problem related to this method which generates random bytes that cannot be converted to a valid hash hex

lesterli: After an offline discussion, this method is fine. Our previous understanding is wrong, the generated random bytes can be converted to a valid hex hash, like`hex.EncodeToString(testBlock.Hash)`.

PR #438, opened 2024-06-27
fix: commit pub rand issue

lesterli:
## Summary

This PR fixes the pub rand issue which when `NumPubRand` is 0, the integer underflow can happen. This is the same problem https://github.com/babylonchain/babylon-contract/issues/186

## Test Plan

```
make test
make test-e2e
```

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -897,8 +897,12 @@ func (fp *FinalityProviderInstance) GetLastCommittedHeight() (uint64, error) {
 	}
 
 	// no committed randomness yet
-	if pubRandCommit == nil {
+	if pubRandCommit == nil || (pubRandCommit.StartHeight == 0 && pubRandCommit.NumPubRand == 0) {
 		return 0, nil
+	}
+
+	if pubRandCommit.NumPubRand == 0 {
+		return 0, errors.New("the field NumPubRand should always be at least one")
 	}
 
 	lastCommittedHeight := pubRandCommit.StartHeight + pubRandCommit.NumPubRand - 1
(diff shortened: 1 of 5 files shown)


bap2pecs (approved):

finality-provider/service/fp_instance.go
```diff
@@ -897,10 +897,14 @@ func (fp *FinalityProviderInstance) GetLastCommittedHeight() (uint64, error) {
 	}
 
 	// no committed randomness yet
-	if pubRandCommit == nil {
+	if pubRandCommit == nil || (pubRandCommit.StartHeight == 0 && pubRandCommit.NumPubRand == 0) {
```
SebastianElvis, line 900: In which case `pubRandCommit` is not nil but empty? If this is ever possible then how about we fix `lastCommittedPublicRandWithRetry` or the function in the client controller instead?

lesterli: When there is no committed randomness yet, it returns empty from `QueryLastCommittedPublicRand`. 
Maybe we can update it and ensure it always returns `nil` rather than the empty value. @bap2pecs Wdyt?

bap2pecs: @SebastianElvis it's actually more complicated. if anyone implemented `QueryLastCommittedPublicRand` wrong and returns empty struct, this can happen.

I think we can still have this logic to be extra safe. or even better, instead of `return 0, nil`, we should return error.

SebastianElvis: > I think we can still have this logic to be extra safe. 

Sounds good to me. Indeed other impl of this interface might make such mistake

> or even better, instead of return 0, nil, we should return error.

Do not recommend this one as it's not essentially an error, but rather the FP hasn't committed pub rand yet

bap2pecs: cc @gitferry

gitferry: Maybe we can have `commit.Validate()` in both client controller impl and it's caller for extra safe?

bap2pecs: i like this idea. i will create another PR for this

PR #441, opened 2024-06-27
chore: update op e2e test

lesterli:
## Summary

This PR does the following updates for op e2e test:

* make the block hash random and hex encoding it when querying
* update the rand source with `time.Now().Unix()`

## Test Plan

```
make test
make test-e2e
```


bap2pecs:
to add some more context, we made the change initially based on a wrong assumption but the root cause was not this so we now change it back

bap2pecs (approved):

lesterli:
> to add some more context, we made the change initially based on a wrong assumption but the root cause was not this so we now change it back

Indeed, there is the context in the previous pr, https://github.com/babylonchain/finality-provider/pull/437

Issue #390, opened 2024-06-20
OP consumer: QueryFinalityProviderVotingPower implementation


bap2pecs:
per my discussion w @SebastianElvis, we can just hardcode to a positive value b/c so this interface function is only used for checking if the FP is eligible for submitting sigs

but it's still useful. at least honest FPs won't submit useless data

so post-mvp we can do this per @SebastianElvis 's suggestion

> A possible algorithm is as follows: instead of querying the exact voting power, the FP queries BTC delegations page by page, and if there is >=1 active BTC delegation, then it can vote.

> Another way is that FP leverages backend service that indexes voting power for Babylon. This introduces some unnecesasry coupling though

we can think more later

SebastianElvis:
QueryFinalityProviderVotingPower sounds good to me

The proposed QueryActivatedHeight might not work as the start_height is not necessarily the height that BTC delegation becomes active. To become active, it needs to satisfy 1) the staking tx is k-deep in Bitcoin, i.e., start_height + k and 2) it receives a quorum number of covenant committee signatures

bap2pecs:
note that for `QueryFinalityProviderVotingPower`, one problem is if we always return 1, it will cause issue

```
2024-06-29T17:15:40.891-0400	FATAL	service/fp_manager.go:110	terminating the finality-provider instance due to critical error	{"pk": "12b1058e00115ea41f28582db7fb684c03113933c630ffbfb5baea91beec6cc1", "error": "reached max failed cycles with err: failed to get inclusion proof of public randomness 983078c40dea8dfb765005d207092437f867bc881e1e45220cf15bfff2b056cd for FP 12b1058e00115ea41f28582db7fb684c03113933c630ffbfb5baea91beec6cc1 for block 2: public randomness proof not found"}
github.com/babylonchain/finality-provider/finality-provider/service.(*FinalityProviderManager).monitorCriticalErr
	/Users/zidong/Documents/Projects/babylon-finality-provider/finality-provider/service/fp_manager.go:110
FAIL	github.com/babylonchain/finality-provider/itest/opstackl2	58.684s
FAIL
```

due to submission loop trying to submit sig for a height missing PR

PR #447, opened 2024-06-28
test: spin up OP devnet in e2e test

bap2pecs:
## Summary

<details><summary>added optimism codebase as submodule</summary>
b/c we want to reuse some infra to set up an OP devnet.

Note: not importing the upstream b/c we don't want to depend on both our forked OP codebase and the official one at the same time
</details>

## Test Plan


go.mod
```diff
@@ -310,7 +431,12 @@ require (
 replace (
 	// use cosmos fork of keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
-	github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.8.6-0.20240618010559-dc7b6269474d
+	// github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.8.6-0.20240618010559-dc7b6269474d
+	github.com/babylonchain/babylon => ../babylon-private
+	github.com/cockroachdb/pebble => github.com/cockroachdb/pebble v0.0.0-20231018212520-f6cde3fc2fa4
+	// github.com/ethereum-optimism/optimism => github.com/babylonchain/optimism 9f823ab3ed49c31d9af1f589946e76ae7b2c4e9f
+	github.com/ethereum-optimism/optimism => ../babylon-optimism
```
bap2pecs, line 438: will fix before merging

clientcontroller/opstackl2/consumer.go
```diff
@@ -34,15 +34,17 @@ var _ api.ConsumerController = &OPStackL2ConsumerController{}
 
 type OPStackL2ConsumerController struct {
 	Cfg        *fpcfg.OPStackL2Config
+	fpCfg      *fpcfg.Config
```
bap2pecs, line 37: added b/c we need to know NumPubRand in `QueryFinalityProviderVotingPower`

clientcontroller/opstackl2/consumer.go
```diff
@@ -223,10 +226,34 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
...
+		return 0, fmt.Errorf("failed to query smart contract state: %w", err)
+	}
+
+	var hasPubRandCommit bool
+	err = json.Unmarshal(stateResp.Data, &hasPubRandCommit)
+	if err != nil {
+		return 0, fmt.Errorf("failed to unmarshal response: %w", err)
+	}
+
+	if !hasPubRandCommit {
```
bap2pecs, line 250: need this check b/c of https://github.com/babylonchain/finality-provider/issues/390#issuecomment-2198388270

itest/opstackl2/op_test_manager.go
```diff
@@ -57,14 +59,16 @@ type OpL2ConsumerTestManager struct {
...
+	OpChainId         string
 }
 
 func StartOpL2ConsumerManager(t *testing.T) *OpL2ConsumerTestManager {
 	// Setup consumer test manager
 	testDir, err := e2eutils.BaseDir("fpe2etest")
 	require.NoError(t, err)
 
-	logger := zap.NewNop()
+	logger := createLogger(t, zapcore.ErrorLevel)
```
bap2pecs, line 71: no-op logger made debugging so hard. see https://github.com/babylonchain/finality-provider/issues/449

I am using `ErrorLevel` here which won't flood the output stream while still give me important logs I need to see

clientcontroller/opstackl2/consumer.go
SebastianElvis: In this case can we remove `Cfg` as `fpCfg` seems to contain `Cfg` already

clientcontroller/opstackl2/consumer.go
```diff
@@ -223,10 +226,34 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
 }
 
 // QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height.
-// This interface function only used for checking if the FP is eligible for submitting sigs.
-// Now we can simply hardcode the voting power to a positive value.
-// TODO: see this issue https://github.com/babylonchain/finality-provider/issues/390 for more details
 func (cc *OPStackL2ConsumerController) QueryFinalityProviderVotingPower(fpPk *btcec.PublicKey, blockHeight uint64) (uint64, error) {
```
SebastianElvis, line 229: The function checks whether the FP has committed pub rand at the given height. However, this is not necessarily relevant to whether the FP has voting power at the height. A FP that does not receive any BTC delegation can still commit pub rand. Even post MVP, we should not couple the pub rand commit with the voting power table.

itest/opstackl2/devnet-data/allocs-l1.json
SebastianElvis, line 1: Those json files are super big (50000LoCs!). Why do they need to be in the code base, can they be git ignored? For example, can we generate those json or retrieve them somewhere upon running e2e tests?

SebastianElvis (commented):
great work glueing everything together! some initial comment

clientcontroller/opstackl2/consumer.go
bap2pecs: agree that coupling is not good. but hardcoding it to always return 1 will result in FP trying to submit sig where there is no pub rand. and the FP will stuck in the submission loop and fatal.

I think the problem is this function probably should be renamed to sth like CanSubmitSigAtHeight()

wdyt?

itest/opstackl2/devnet-data/allocs-l1.json
bap2pecs: this is generated by running `make devnet-allocs` inside https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e

this is needed before running e2e test to set up the op system.

it's generated by this local python code https://github.com/ethereum-optimism/optimism/tree/develop/bedrock-devnet/devnet so unfortunately can't be used in our e2e test. importing the entire op codebase as a submodule is also not a good idea (we tried but it makes our repo a mess due to lots of recursive submodules)

so far this seems to be the cleanest way but happy to hear other suggestions

clientcontroller/factory.go
```diff
@@ -41,7 +41,7 @@ func NewConsumerController(config *fpcfg.Config, logger *zap.Logger) (api.Consum
 			return nil, fmt.Errorf("failed to create Babylon rpc client: %w", err)
 		}
 	case OPStackL2ConsumerChainName:
-		ccc, err = opstackl2.NewOPStackL2ConsumerController(config.OPStackL2Config, logger)
+		ccc, err = opstackl2.NewOPStackL2ConsumerController(config, logger)
```
gitferry, line 44: Seems only `fpConfig.OPStackL2Config` is used here

clientcontroller/opstackl2/consumer.go
gitferry: We have a separate [check](https://github.com/babylonchain/finality-provider/blob/01dbfb5514e9e938792bff1d2636ba288456c159/finality-provider/service/fastsync.go#L66) regarding whether pub rand is committed. Why do we need to couple them here?

finality-provider/service/fp_instance.go
```diff
@@ -725,7 +725,13 @@ func (fp *FinalityProviderInstance) SubmitFinalitySignature(b *types.BlockInfo)
 	// get inclusion proof
 	proofBytes, err := fp.pubRandState.GetPubRandProof(pubRand)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get inclusion proof of public randomness: %v", err)
+		return nil, fmt.Errorf(
+			"failed to get inclusion proof of public randomness %s for FP %s for block %d: %v",
```
gitferry, line 729:
```suggestion
			"failed to get inclusion proof of public randomness %s for FP %s for block %d: %w",
```
nit: error can better be wrapped

gitferry (commented):
Great work! Some initial comments:

clientcontroller/opstackl2/consumer.go
bap2pecs: the problem is `GetLastCommittedHeight()` returns height 0 if there is no committed randomness yet

so `hasRandomness()` will always return `true` per implementation here:

```
func (fp *FinalityProviderInstance) hasRandomness(b *types.BlockInfo) (bool, error) {
	lastCommittedHeight, err := fp.GetLastCommittedHeight()
	if err != nil {
		return false, err
	}
	if b.Height > lastCommittedHeight {
		fp.logger.Debug(
			"the finality provider has not committed public randomness for the height",
			zap.String("pk", fp.GetBtcPkHex()),
			zap.Uint64("block_height", b.Height),
			zap.Uint64("last_committed_height", lastCommittedHeight),
		)
		return false, nil
	}

	return true, nil
}
```

gitferry: if `GetLastCommittedHeight` returns 0, then `b.Height > lastCommittedHeight` would be true as block height is always `>=1`, then `false` will be returned. Am I missing anything?

clientcontroller/factory.go
bap2pecs: good catch. since I have updated https://github.com/babylonchain/babylon-contract/pull/196 this afternoon, we no longer need to pass in NumPubRand. so I just reverted this change here as well

itest/opstackl2/devnet-data/allocs-l1.json
gitferry: Can we put the static file in the other repo where is responsible for such files?

clientcontroller/opstackl2/consumer.go
bap2pecs: you are right. my bad. the problem is not `hasRandomness`

itest/opstackl2/devnet-data/allocs-l1.json
bap2pecs: we can do that. but why is that better than keeping it in this repo where the e2e tests are located? imo it's similar to that we put the CW binary files under `itest/bytecode`

gitferry: For users who just want to download the repo and run the binaries, they don't need to download these static files that are just for testing purpose

bap2pecs: @gitferry that makes sense. could you create a new repo (public visibility) for storing those static files and add me and @lesterli?

gitferry: let's keep them in this repo for now. Will need @SebastianElvis's input as storing static files might not be specific to OP L2

clientcontroller/opstackl2/msg.go
```diff
@@ -38,6 +38,7 @@ type SubmitFinalitySignatureResponse struct {
 type QueryMsg struct {
 	Config            *Config            `json:"config,omitempty"`
 	LastPubRandCommit *LastPubRandCommit `json:"last_pub_rand_commit,omitempty"`
+	HasPubRandCommit  *HasPubRandCommit  `json:"has_pub_rand_commit,omitempty"`
```
maurolacy, line 41: Nit: Rename these to `PubRandCommit` for clarity.

clientcontroller/opstackl2/msg.go
```diff
@@ -46,6 +47,12 @@ type LastPubRandCommit struct {
 	BtcPkHex string `json:"btc_pk_hex"`
 }
 
+type HasPubRandCommit struct {
+	BtcPkHex   string `json:"btc_pk_hex"`
+	Height     uint64 `json:"height"`
+	NumPubRand uint64 `json:"num_pub_rand"`
```
maurolacy, line 53: I don't think this field is part of the query.

clientcontroller/opstackl2/consumer.go
```diff
@@ -223,10 +223,33 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
...
+	if err != nil {
+		return 0, fmt.Errorf("failed marshaling to JSON: %w", err)
+	}
+	stateResp, err := cc.CwClient.QuerySmartContractState(cc.Cfg.OPFinalityGadgetAddress, string(jsonData))
+	if err != nil {
+		return 0, fmt.Errorf("failed to query smart contract state: %w", err)
+	}
+
+	var hasPubRandCommit bool
+	err = json.Unmarshal(stateResp.Data, &hasPubRandCommit)
```
maurolacy, line 241: Funny way to check for success. Perhaps it's better to deserialise the thing and see if it's valid.

itest/bytecode/op_finality_gadget_fc754df.wasm
maurolacy, line 1: Nit: Use a fixed name here, and put the commit id / version tag into a `version.txt` file instead.

itest/opstackl2/op_test_manager.go
```diff
@@ -33,15 +34,16 @@ import (
...
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 const (
-	opFinalityGadgetContractPath = "../bytecode/op_finality_gadget_1947cc6.wasm"
-	opConsumerId                 = "op-stack-l2-12345"
-	// it has to be a valid EVM RPC which doesn't timeout
-	opStackL2RPCAddress = "https://eth.drpc.org"
+	opFinalityGadgetContractPath = "../bytecode/op_finality_gadget_fc754df.wasm"
```
maurolacy, line 46: Nit: Use fixed name.

itest/opstackl2/op_test_manager.go
maurolacy, line 71: Nit: The logger-related changes would have been better living in another PR.

maurolacy (approved):
LGTM, AFAIK. The Eth L1 / L2 stuff / setup eludes me.

itest/opstackl2/devnet-data/allocs-l1.json
SebastianElvis: > it's generated by this local python code [ethereum-optimism/optimism@develop/bedrock-devnet/devnet](https://github.com/ethereum-optimism/optimism/tree/develop/bedrock-devnet/devnet?rgh-link-date=2024-06-30T13%3A48%3A15Z) so unfortunately can't be used in our e2e test. importing the entire op codebase as a submodule is also not a good idea (we tried but it makes our repo a mess due to lots of recursive submodules)

Is it possible that in the Makefile entry, we do `git clone` to clone optimism code base and wire up some bash scripts such that the e2e test can grab these json files and spin up the devnet?

Tbh it still a bit itchy that one needs to clone so much data to run the finality provider, and the data is solely for testing purposes

bap2pecs: yeah I checked the data files they are about 30M is total so agree that it's probably not a good idea to keep ti here considering the entire repo is 300MB only

> we do git clone to clone optimism code base and wire up some bash scripts such that the e2e test can grab these json files and spin up the devnet?

it doesn't exists in the OP's repo. it's generated by running `make devnet-allocs` and the files are not even under one place. 5 files will be generated under `.devnet/` and 1 file will be generated under `packages/contracts-bedrock/deploy-config/devnetL1.json`. so we have to collect those files from these two places.

so I'd rather prefer not to generate these files on the fly b/c: 1) clone, generate and collet those files adds unnecessary complexities in our codebase; 2) it also increases e2e test running time further

if the concern that we don't want to create a new repo under babylon, we can use aws s3 bucket to store those files easily. only downside is we lose version control

itest/opstackl2/op_test_manager.go
bap2pecs: good point. I just moved the code to https://github.com/babylonchain/finality-provider/pull/451

clientcontroller/opstackl2/consumer.go
bap2pecs: isn't `Unmarshal` the deserialise process? not sure if I am missing anything

itest/opstackl2/devnet-data/allocs-l1.json
SebastianElvis: created https://github.com/babylonchain/op-e2e-devnet as per offline discussion

bap2pecs: ok I just added a command `make op-e2e-devnet` in CI

itest/bytecode/op_finality_gadget_fc754df.wasm
maurolacy: Your call. We use fixed filenames (and a couple download / copy scripts) everywhere.

You can always check the integrity of the wasm using `sha256sum`, and comparing to the original. A bad rebase can hit you anyway, by the way. By example on the file name references.

clientcontroller/opstackl2/consumer.go
bap2pecs: I think that's due to the API return type. we hv sth similar here as well: https://github.com/babylonchain/babylon-da-sdk/blob/main/sdk/sdk.go#L289-L292

but since we have #454 now, I removed all the changes in this `QueryFinalityProviderVotingPower` method

itest/bytecode/op_finality_gadget_fc754df.wasm
bap2pecs: yes we were hit by that once and spent some time figuring it out. but let's do that in a separate PR. just created https://github.com/babylonchain/finality-provider/issues/457 to track

bap2pecs:
ok all tests passed

PR #448, opened 2024-06-29
test: create anvil devnet for tests

parketh:
## Summary

Current OP e2e tests use an external RPC provider which is subject to rate limiting.

This PR replaces the external RPC with a local [Anvil](https://book.getfoundry.sh/anvil/) devnet instance run inside the test manager. It is run in memory and torn down at test completion, with logs stored at `anvil.log` for debugging. 

## Test Plan

```
make install
make test e2e-op
```

Note: this PR is superseded by https://github.com/babylonchain/finality-provider/pull/447

### DIFF
--- a/itest/ethereum_node_handler.go
+++ b/itest/ethereum_node_handler.go
@@ -1,0 +1,101 @@
+package e2e_utils
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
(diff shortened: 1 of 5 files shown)


bap2pecs:
yeah let's keep this one. it will still be useful if we add other non-OP chains

Issue #439, opened 2024-06-27
flaky `make test`


bap2pecs:
[example](https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1589/workflows/9e581924-d3e1-404a-89a8-83cd29a16658/jobs/3657/parallel-runs/0/steps/0-115)

![image](https://github.com/babylonchain/finality-provider/assets/111917526/64c81685-0aee-4cb5-9b0f-5d75b3895892)

this happens a bit too frequent cc @SebastianElvis

SebastianElvis:
i guess it's simply that we have too many e2e tests so that it sometimes time out? https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1589/workflows/9e581924-d3e1-404a-89a8-83cd29a16658/jobs/3657/parallel-runs/0/steps/0-115?invite=true#step-115-2974_33

If yes, how about we break down the CI pipeline a bit, e.g., 1 workflow for unit tests, 1 workflow for Babylon e2e, 1 workflow for wasm e2e, and 1 workflow for OP e2e?

bap2pecs:
@SebastianElvis this is before the `make test-e2e` step. so I don't think the timeout is caused by e2e. if you see this one, you can see normal `make test` only takes 1 minute but this one takes 10 minutes so it timed out

<img width="1681" alt="image" src="https://github.com/babylonchain/finality-provider/assets/111917526/ac93f715-2c03-4301-b9f5-1d9b56ff91c8">

so I think we need to investigate into the unit test to understand why it failed sometimes. maybe it's related to the fuzzing tests but I am not sure

> 1 workflow for Babylon e2e, 1 workflow for wasm e2e, and 1 workflow for OP e2e

but this is a good idea. we can do it

SebastianElvis:
Hmm I think you are right.

Digging into the error stack and I see

```
goroutine 222 [chan receive, 9 minutes]:
testing.(*F).Fuzz.func1({0x3c05ca0?, 0xc0017ec5a0}, {{0x0, 0x0}, {0xc000e0b7d8, 0x6}, {0x0, 0x0, 0x0}, {0xc000a42b10, ...}, ...})
        /usr/local/go/src/testing/fuzz.go:337 +0x5aa
testing.(*F).Fuzz(0xc0017ec5a0, {0x2d51ee0, 0x367be20})
        /usr/local/go/src/testing/fuzz.go:402 +0xa37
[...]

PR #452, opened 2024-07-01
fix: add missing Config.Validate()

bap2pecs:
## Summary

just realized that we have define this method in `finality-provider/config/opstackl2.go` but never used it

not sure why but let's add it

## Test Plan

```
make lint
```
wait for CI


gitferry (approved):
CI seems flacky

bap2pecs:
> CI seems flacky

seems to be the same as testhttps://github.com/babylonchain/finality-provider/issues/439

cc @gitferry

PR #453, opened 2024-07-01
style: fix long lines using golines

bap2pecs:
## Summary

long lines are hard to read and go fmt doesn't fix it for us. this PR is generated purely by running `golines .  -w` to wrap lines at 120 length

unfortunately we can't enforce it in `.golangci.yml` by adding

```
linters:
  enable:
    - lll
```

due to https://github.com/golangci/golangci-lint/issues/3983

but we can probably add a check like: https://github.com/babylonchain/finality-provider/pull/352/files 

## Test Plan

```
make lint
```
wait for CI

### DIFF
--- a/clientcontroller/api/interface.go
+++ b/clientcontroller/api/interface.go
@@ -34,13 +34,31 @@ type ClientController interface {
 type ConsumerController interface {
 	// CommitPubRandList commits a list of EOTS public randomness the consumer chain
 	// it returns tx hash and error
-	CommitPubRandList(fpPk *btcec.PublicKey, startHeight uint64, numPubRand uint64, commitment []byte, sig *schnorr.Signature) (*types.TxResponse, error)
+	CommitPubRandList(
+		fpPk *btcec.PublicKey,
+		startHeight uint64,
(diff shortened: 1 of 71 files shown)


SebastianElvis:
Not sure if we should use multiple linters for the same project, as those linters might give comflicting suggestions. Let's see others' opinions as well.

Also, if the concept is ACK'ed, could we submit PR to the public `dev` branch rather than the base branch?

bap2pecs:
> Also, if the concept is ACK'ed, could we submit PR to the public `dev` branch rather than the base branch?

yes I am only doing it as a RFC first :)

maurolacy:
This will add an extra pain to the rebasing / merging from the main `dev` branch process.
I don't think we should be doing this. At least, not at this point in the dev cycle.

maurolacy (commented):
I propose we don't do this now. Not even against the main `dev` branch.

bap2pecs:
agree that it's better to hold onto it. after we consolidate the two branches would be a better timing

maurolacy:
Closing / delayed until base and dev are merged.

Issue #450, opened 2024-06-30
op-e2e: BBN FP shouldn't use the OP consumer to submit PR and sigs


bap2pecs:
found this weird behavior when debugging. root cause is b/c it shares the same fpApp with OP FPs and the fpApp is configured to use `opcc`:

```
fpApp, err := service.NewFinalityProviderApp(cfg, bc, opcc, eotsCli, fpdb, logger)
```

this doesn't break anything but I am worried that when the tests get complicated it might cause some very hard to debug issues in the future

bap2pecs:
we should figure out a way to set it up properly. another way is to remove the restriction in tests that require a delegation must also delegate to BBN FP. we can add a boolean in the [here](https://github.com/babylonchain/babylon-private/blob/base/consumer-chain-support/x/btcstaking/keeper/btc_delegations.go#L211)

I think although it's a business decision, we can still make our life easier by removing this restriction in e2e test. as I imagine when we add more consumers in the future, it's going to be too complicated to maintain all the FPs in e2e tests

wdyt @SebastianElvis

SebastianElvis:
The finality provider implementation does not allow maintaining FP instances from different consumers. So the e2e test might need to be restructured. I see the following possibilities:

- (preferred) have an admin in the Babylon OP contract, use the admin to inject voting power for the FP, such that FP can submit things to the OP contract
- manually wire a Babylon FP registration request and send to Babylon

CosmWasm e2e uses the admin account approach. Wdyt?

bap2pecs:
> have an admin in the Babylon OP contract, use the admin to inject voting power for the FP, such that FP can submit things to the OP contract. This approach does not involve delegation at all

the doesn't work in our case because we need to query the actual delegation here: https://github.com/babylonchain/babylon-da-sdk/blob/main/sdk/sdk.go#L229-L258

and our current test case already tests that this logic is correct. I think it's good to keep the real delegation in e2e tests to test logic like this since there are so many moving parts in the system so injection should be used carefully

> manually wire a Babylon FP registration request and send to Babylon

yes that's what we are doing in our test but tbh it's not relevant to our code. it's only to satisfy:

```
	if !restakedToBabylon {
		// a BTC delegation has to stake to at least 1 Babylon finality provider
		return false, types.ErrNoBabylonFPRestaked
	}
```

maybe we can build some e2e test infra to make it less confusing

bap2pecs:
[...]

PR #454, opened 2024-07-01
fix: start height of initial public randomness commit

bap2pecs:
## Summary

see discussion here https://github.com/babylonchain/finality-provider/pull/447#discussion_r1660074723

let's first list the facts and requirements:

facts:
- first PR submissoin is based on tip height (e.g. height 27)
- chain poller starts from the lowest non-finalized height that the fp has not processed. (e.g. height 2)
- finalitySigSubmissionLoop receives signal from chain poller to try to submit sig from height 2 which will ofc fatal on max retry

requirements:
- we want to ensure no blocks with voting power will be missed

---

so I think the root cause is first PR submission shouldn't be determined by tipHeight. b/c it means the PR and Sig submission loops is not consistent in the start state.

my proposed fix is to add a global state firstPolledHeight that only be set once on the first poll to trigger the first PR submission

## Test Plan

comment out our changes in QueryFinalityProviderVotingPower first

```
make test-e2e-op
```

see test passed
```
$ make test-e2e-op                                                                    [15:22:07]
No processes to kill
rm -rf ~/.babylond ~/.wasmd ~/.bcd
cd tools; \
	go install -trimpath github.com/babylonchain/babylon/cmd/babylond
go test -mod=readonly -timeout=25m -v github.com/babylonchain/finality-provider/itest/opstackl2 -count=1 --tags=e2e_op --run ^TestBlockBabylonFinalized
=== RUN   TestBlockBabylonFinalized
    babylon_node_handler.go:148: babylon log file: /var/folders/9_/q4wsdnh14_s60_74cd2rbztm0000gp/T/zBabylonTest1670807382/babylon.log
    op_test_manager.go:83: Fp home dir: /var/folders/9_/q4wsdnh14_s60_74cd2rbztm0000gp/T/fpe2etest1412267153/fp-home
    setup.go:490: Generating L2 genesis l2_allocs_mode delta
    op_test_manager.go:109: Register consumer op-stack-l2-901 to Babylon
    op_test_manager.go:153: Deployed op finality contract address: bbn14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sw76fy2
    op_test_manager.go:241: Babylon node has started
    op_test_manager.go:339: Registered Finality Provider b3996108efd9c82ca02dcc521fb39e20c2278ebf9ab58e9382b7ed962df9f64f for chain-test
    op_test_manager.go:383: The test manager is running with 1 finality-provider(s)
    op_test_manager.go:339: Registered Finality Provider 1c4b9c180ad2e9aa832e593490fa1d19e0b5fc7e81a68f12cd30efdcf12e7990 for op-stack-l2-901
    op_test_manager.go:339: Registered Finality Provider 08f021d4433b713c638fbe150f3446932ef01263de130db32ade5815b18eaba2 for op-stack-l2-901
    op_test_manager.go:383: The test manager is running with 3 finality-provider(s)
    utils.go:101: Public randomness for fp 08f021d4433b713c638fbe150f3446932ef01263de130db32ade5815b18eaba2 is successfully committed at height 65
    utils.go:101: Public randomness for fp 08f021d4433b713c638fbe150f3446932ef01263de130db32ade5815b18eaba2 is successfully committed at height 129
    utils.go:101: Public randomness for fp 1c4b9c180ad2e9aa832e593490fa1d19e0b5fc7e81a68f12cd30efdcf12e7990 is successfully committed at height 193
    utils.go:101: Public randomness for fp 1c4b9c180ad2e9aa832e593490fa1d19e0b5fc7e81a68f12cd30efdcf12e7990 is successfully committed at height 257
    base_test_manager.go:174: successfully submitted a BTC delegation
    base_test_manager.go:174: successfully submitted a BTC delegation
    base_test_manager.go:207: delegations are pending
    base_test_manager.go:227: delegations are active
    op_test_manager.go:284: Test block height 706 and 706
    op_test_manager.go:314: Submit finality signature to op finality contract &{Height:706 Hash:[163 198 161 192 245 28 28 32 201 102 21 235 1 197 250 177 144 130 99 49 168 224 213 83 0 193 77 184 49 152 223 82]}
    op_test_manager.go:314: Submit finality signature to op finality contract &{Height:706 Hash:[163 198 161 192 245 28 28 32 201 102 21 235 1 197 250 177 144 130 99 49 168 224 213 83 0 193 77 184 49 152 223 82]}
    op_e2e_test.go:190: Test case 1: block 706 is finalized
    op_test_manager.go:314: Submit finality signature to op finality contract &{Height:707 Hash:[185 213 238 184 77 103 140 11 128 213 132 155 49 115 2 116 83 205 111 62 204 148 199 207 229 103 136 55 144 13 147 230]}
    op_e2e_test.go:210: Test case 2: block 707 is not finalized
    op_test_manager.go:464: Stopping test manager
--- PASS: TestBlockBabylonFinalized (254.95s)
PASS
ok  	github.com/babylonchain/finality-provider/itest/opstackl2	256.528s
```

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -119,7 +119,7 @@ func (fp *FinalityProviderInstance) Start() error {
 
 	poller := NewChainPoller(fp.logger, fp.cfg.PollerConfig, fp.cc, fp.consumerCon, fp.metrics)
 
-	if err := poller.Start(startHeight + 1); err != nil {
+	if err := poller.Start(startHeight); err != nil {
 		return fmt.Errorf("failed to start the poller: %w", err)
 	}
 
@@ -132,7 +132,7 @@ func (fp *FinalityProviderInstance) Start() error {
 	fp.wg.Add(1)
 	go fp.finalitySigSubmissionLoop()
 	fp.wg.Add(1)
-	go fp.randomnessCommitmentLoop()
+	go fp.randomnessCommitmentLoop(startHeight)
 	fp.wg.Add(1)
 	go fp.checkLaggingLoop()
 
@@ -283,21 +283,50 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
 	}
 }
 
-func (fp *FinalityProviderInstance) randomnessCommitmentLoop() {
+func (fp *FinalityProviderInstance) randomnessCommitmentLoop(startHeight uint64) {
 	defer fp.wg.Done()
 
 	commitRandTicker := time.NewTicker(fp.cfg.RandomnessCommitInterval)
 	defer commitRandTicker.Stop()
 
+	lastCommittedHeight, err := fp.GetLastCommittedHeight()
+	if err != nil {
+		fp.logger.Fatal("Error getting last committed height while starting the randomness commitment loop", zap.Error(err))
+		return
+	}
+
+	// if there is no committed randomness, we need to commit the first randomness
+	if lastCommittedHeight == uint64(0) {
+		txRes, err := fp.retryCommitPubRandUntilMaxRetry(startHeight)
+		if err != nil {
+			fp.metrics.IncrementFpTotalFailedRandomness(fp.GetBtcPkHex())
+			fp.reportCriticalErr(err)
+			return
+		}
+		if txRes == nil {
+			fp.logger.Fatal(
+				"Error submitting the first randomness",
+				zap.String("consumer_id", string(fp.GetChainID())),
+				zap.String("pk", fp.GetBtcPkHex()),
+			)
+			return
+		}
+		fp.logger.Info(
+			"successfully committed public randomness to the consumer chain",
+			zap.String("consumer_id", string(fp.GetChainID())),
+			zap.String("pk", fp.GetBtcPkHex()),
+		)
+	}
+
 	for {
 		select {
 		case <-commitRandTicker.C:
-			tipBlock, err := fp.getLatestBlockHeightWithRetry()
+			tipBlockHeight, err := fp.getLatestBlockHeightWithRetry()
 			if err != nil {
 				fp.reportCriticalErr(err)
 				continue
 			}
-			txRes, err := fp.retryCommitPubRandUntilBlockFinalized(tipBlock)
+			txRes, err := fp.retryCommitPubRandUntilBlockFinalized(tipBlockHeight)
 			if err != nil {
 				fp.metrics.IncrementFpTotalFailedRandomness(fp.GetBtcPkHex())
 				fp.reportCriticalErr(err)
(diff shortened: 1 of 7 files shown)


bap2pecs:
will fix test after getting some feedback on this approach

bap2pecs:
just had a discussion w @gitferry offline. the problem of this issue is there is a subtle assumption: when a FP first started, it doesn't have any voting power yet.

given that when we add timestamping later, the entire logic here for handling FP startup will need a massive refactoring, it's not worth it to make this temporary fix.

for now, we can just modify our e2e tests to satisfy this assumption

bap2pecs:
@gitferry and I discussed again and decided to fix it before the timestamping feature is added.

finality-provider/service/fp_instance.go
```diff
@@ -283,21 +283,50 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+			return
+		}
+		if txRes == nil {
+			fp.logger.Fatal(
+				"Error submitting the first randomness",
+				zap.String("consumer_id", string(fp.GetChainID())),
+				zap.String("pk", fp.GetBtcPkHex()),
+			)
+			return
+		}
```
gitferry, line 313: Is this check necessary? This case can happen when there's already sufficient randomness. Besides, `txRes` is not used in other places

finality-provider/service/fp_instance.go
```diff
@@ -883,17 +956,14 @@ func (fp *FinalityProviderInstance) getPollerStartingHeight() (uint64, error) {
 	if err != nil {
```
gitferry, line 956: the doc string above can be modified

gitferry (approved):
Great fix! Some nits:

finality-provider/service/fp_instance.go
bap2pecs: > This case can happen when there's already sufficient randomness

that's for the other method i.e. `retrySubmitFinalitySignatureUntilBlockFinalized`

in this one, it should never happen b/c we checked `lastCommittedHeight == uint64(0)` above. but I still want to have this check to be safe b/c `retryCommitPubRandUntilMaxRetry` can return `nil`

this is due to `CommitPubRand` can return `nil`. but actually I think it's better to add the check inside `retryCommitPubRandUntilMaxRetry`. let me fix it

> Besides, txRes is not used in other places

good catch. let me remove it

bap2pecs: actually, it doesn't make sense to add the check inside `retryCommitPubRandUntilMaxRetry` b/c we shouldn't assume the input argument is always for the case when there is no PR committed yet.

there are only two cases where both `txRes` and `err` will be `nil`

1) when there is sufficient PR. not that's not right b/c when this code is hit, there is no PR at all. so something very bad must have happened
2) when `CommitPubRandList()` returns `nil, nil`. this is not right b/c if there is no error, we should expect the `TxResponse` not to be `nil`

so if we need to check `txRes`, we cannot remove it

bap2pecs:
```
fp_instance.go:771: Unexpected call to *mocks.MockConsumerController.CommitPubRandList([0xc0000853b0 26 100 [47 103 60 142 28 203 22 171 80 137 209 89 214 181 20 214 208 28 246 52 54 180 112 52 24 68 224 145 195 219 12 131] 0xc0022aa9b0]) at /home/circleci/project/finality-provider/service/fp_instance.go:771 because: there are no expected calls of the method "CommitPubRandList" for that receiver
    --- FAIL: FuzzRegisterFinalityProvider/seed#7 (0.36s)
```

this failure seems relevant

finality-provider/service/fp_instance.go
gitferry: OK then. Let's keep the check

bap2pecs:
ok just fixed a bunch of tests

gitferry:
Thanks for fixing the tests. A kind reminder is to better merge it to `dev` (title could be reworded)

bap2pecs:
@gitferry sure can I cherry-pick it to `dev` later in a separate PR?

gitferry:
> any suggestions?

Something like `fix: start height of initial public randomness commit`?

bap2pecs:
> > any suggestions?
> 
> Something like `fix: start height of initial public randomness commit`?

ah I see you point. the title was indeed inappropriate since now we don't add the extra internal state in the poller

finality-provider/service/fp_instance.go
bap2pecs: jfyi this already detects the same issue we found here https://github.com/babylonchain/finality-provider/issues/450

```
    op_test_manager.go:285: Submit finality signature to op finality contract &{Height:642 Hash:[22 13 54 122 191 151 237 207 57 27 24 204 19 129 71 148 217 144 162 163 102 162 194 9 60 109 71 72 12 74 216 164]}
    op_e2e_test.go:210: Test case 2: block 642 is not finalized
    op_test_manager.go:435: Stopping test manager
2024-07-02T13:26:50.818Z        FATAL   service/fp_instance.go:307      Error submitting the first randomness   {"consumer_id": "chain-test", "pk": "b4209be0ad8cc86c658d3e6ff0971e90926c5869dc36494b5e8066f186434f22"}
github.com/babylonchain/finality-provider/finality-provider/service.(*FinalityProviderInstance).randomnessCommitmentLoop
        /home/circleci/project/finality-provider/service/fp_instance.go:307
FAIL    github.com/babylonchain/finality-provider/itest/opstackl2       337.629s
```

PR #455, opened 2024-07-01
[RFC] chore: enable govet linter

bap2pecs:
## Summary

there was a weird bug in this code that took us lots of time to debug:

image

the root cause is we should use `=` instead of `:=` to avoid shadowing variables

after some research, I think we can use tools such as the 'shadow' analyzer in govet to detect this kind of problem

there are many other govet analyzers to use https://golangci-lint.run/usage/linters/#govet but I am just setting it up as an example

I think it will help improve code quality and avoid bugs. happy to hear any feedback

## Test Plan

```
make lint
```

saw 
```
itest/opstackl2/op_test_manager.go:253:3: shadow: declaration of "blocks" shadows declaration at line 250 (govet)
                blocks, err := ctm.OpL2ConsumerCtrl.QueryBlocks(startHeight, startHeight+uint64(n-1), uint64(n))
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,3 +4,11 @@ run:
     - e2e_babylon
     - e2e_bcd
     - e2e_wasmd
+
+linters-settings:
+  govet:
+      # Disable all analyzers.
+      # Default: false
+      disable-all: true
+      enable:
+        - shadow


SebastianElvis:
This is a good move imo. Maybe this can also go to the `dev` branch. @gitferry wdyt?

bap2pecs:
@SebastianElvis note that we needs to fix many places before we can enable this shadow analyzer

as you can see most issues are related to shadowing `err`

image

there is some good discussion here whether it's a good idea to allow shadowing `err`:

https://www.reddit.com/r/golang/comments/se2916/comment/huicxtk/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

SebastianElvis:
Since whether shadow declaration is acceptable is case-by-case, we can always disable it via `//nolint:govet` if necessary

gitferry (commented):
Interesting. Let's move it to `dev` then

maurolacy (approved):
LGTM.

bap2pecs:
just tested to enable all govet linters in the da SDK repo: https://github.com/babylonchain/babylon-da-sdk/pull/38/files#diff-0d98a479cf839de40c441f7803d74816ac06310dfd01b987f33fa3633be5f07f

already found some errors. this seems quite powerful

bap2pecs:
> Interesting. Let's move it to `dev` then

probably better to do it after we consolidate dev and base branches cc @SebastianElvis

SebastianElvis:
Yep let's do it after rebasing base branch

bap2pecs:
> note that we needs to fix many places before we can enable this shadow analyzer

don't think I have the time for this. but I created https://github.com/babylonlabs-io/finality-provider/issues/38 in case anyone wants to pick it up in the future

PR #451, opened 2024-07-01
fix: non-op logger in itests

bap2pecs:
## Summary

https://github.com/babylonchain/finality-provider/issues/449

also per @maurolacy's suggestion https://github.com/babylonchain/finality-provider/pull/447#discussion_r1660667470, I moved the logger changes to this PR

also added some other log related improvements in this PR

## Test Plan

```
make lint
```

wait for CI

PR #456, opened 2024-07-01
chore: improve logs

bap2pecs:
per discussion w @SebastianElvis at https://github.com/babylonchain/finality-provider/pull/451, we should also add those changes in `dev`

### DIFF
--- a/itest/babylon_node_handler.go
+++ b/itest/babylon_node_handler.go
@@ -149,6 +149,7 @@ func NewBabylonNodeHandler(t *testing.T, covenantQuorum int, covenantPks []*type
 
 	f, err := os.Create(filepath.Join(testDir, "babylon.log"))
 	require.NoError(t, err)
+	t.Logf("babylon log file: %s", f.Name())
 
 	startCmd := exec.Command(
 		"babylond",
--- a/finality-provider/config/config.go
+++ b/finality-provider/config/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/jessevdk/go-flags"
+	"go.uber.org/zap/zapcore"
 
 	eotscfg "github.com/babylonchain/finality-provider/eotsmanager/config"
 	"github.com/babylonchain/finality-provider/metrics"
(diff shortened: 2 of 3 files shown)


itest/babylon_node_handler.go
```diff
@@ -149,6 +149,7 @@ func NewBabylonNodeHandler(t *testing.T, covenantQuorum int, covenantPks []*type
 
 	f, err := os.Create(filepath.Join(testDir, "babylon.log"))
 	require.NoError(t, err)
+	t.Logf("babylon log file: %s", f.Name())
```
gitferry, line 152: leftover?

gitferry (approved):

itest/babylon_node_handler.go
bap2pecs: no this is intentionally to print it out as I think this log is helpful for debugging

maurolacy (approved):
LGTM. Thanks for doing this btw.

If this is merged to `dev`, a rebase / merge will be needed to bring it to `base`.

bap2pecs:
@maurolacy actually this is a follow-up PR for #451 which does the same thing so we can save the rebase/merge

maurolacy:
Let's close this one without merging then.

maurolacy:
I'm lost, as this is against dev and the other isn't. Anyway, you figure it out and merge appropiately.

bap2pecs:
@maurolacy yes you can see my discussion with Runchao there. I created this separate PR simply by copying a few changes from that one and apply here.

image

it has to be a separate PR b/c that one has changes related to OP consumer which doesn't exist in `dev`

maurolacy:
Oh, you meant adding the same stuff (improved logging) to both branches. Not ideal (may clash on rebase), but OK for small features I guess.

bap2pecs:
yeah not sure if there is better way for that given that the two branches now diverges a lot

> Not ideal (may clash on rebase), but OK for small features I guess.

wait I thought copying the same code in both branches would save some headaches when we consolidate two branches. or is that not the case?

maurolacy:
> wait I thought copying the same code in both branches would save some headaches when we consolidate two branches. or is that not the case?

Not necessarily. It can lead to conflicts. Anyway, OK for small features / changes.

bap2pecs:
> Not necessarily. It can lead to conflicts.

isn't that similar to `git cherry-pick` a commit? I think this can help reduce merge conflicts when we eventually consolidate the two branches

PR #458, opened 2024-07-02
chore: split CI tests

bap2pecs:
## Summary

per discussion w @SebastianElvis last time, we should split the tests since now it took over 20 mins to run everything

## Test Plan

wait for CI to finish

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,10 +60,98 @@ jobs:
           name: Run tests
           command: |
             make test
+
+  test_e2e_babylon:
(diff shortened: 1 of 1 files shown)


gitferry (approved):

bap2pecs:
ok now it's working.

image

let's see how long it takes

bap2pecs:
ok we cut the CI time to half. seems the bottleneck is `test_e2e_babylon` cc @gitferry @SebastianElvis but this is good for now. we can optimize that one later

image

PR #459, opened 2024-07-02
test: op e2e test with the L2 block

lesterli:
## Summary

This PR updates the op e2e test cases:
* fetch the block from L2 RPC to replace the mock block 
* move the batch submission to a new test case `TestOpSubmitBatchFinalitySigs`

## Test Plan

```
make lint
make test-e2e-op
```


clientcontroller/opstackl2/consumer.go
```diff
@@ -367,6 +367,43 @@ func (cc *OPStackL2ConsumerController) QueryLastPublicRandCommit(fpPk *btcec.Pub
 	return resp, nil
 }
 
+func (cc *OPStackL2ConsumerController) QueryFirstPublicRandCommit(fpPk *btcec.PublicKey) (*types.PubRandCommit, error) {
```
bap2pecs, line 370: I don't this we should add a test-only helper function in production code.

it can also cause confusion that people may think `QueryFirstPublicRandCommit` is part of the interface method

itest/opstackl2/op_test_manager.go
```diff
@@ -240,12 +241,42 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
...
+}
+
 func (ctm *OpL2ConsumerTestManager) WaitForTargetBlockPubRand(t *testing.T, fpList []*service.FinalityProviderInstance, requiredBlockOverlapLen uint64) []*uint64 {
 	require.Equal(t, 2, len(fpList), "The below algorithm only supports two FPs")
 	fpStartHeightList := make([]*uint64, 2)
 	require.Eventually(t, func() bool {
-		firstFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryLastPublicRandCommit(fpList[0].GetBtcPk())
-		secondFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryLastPublicRandCommit(fpList[1].GetBtcPk())
+		firstFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryFirstPublicRandCommit(fpList[0].GetBtcPk())
+		secondFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryFirstPublicRandCommit(fpList[1].GetBtcPk())
```
bap2pecs, line 279: this logic seems incorrect. we don't expect `QueryFirstPublicRandCommit()` to return different value once it's first PR is committed

itest/opstackl2/op_e2e_test.go
```diff
@@ -27,8 +24,7 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	fpList := ctm.StartFinalityProvider(t, false, 1)
 	fpInstance := fpList[0]
 
-	// expect 2 rounds of submissions
-	e2eutils.WaitForFpPubRandCommitted(t, fpInstance, 2)
+	e2eutils.WaitForFpPubRandCommitted(t, fpInstance, 1)
```
bap2pecs, line 27: why change it from to 1?

also, if we will always pass in 1, we should get rid of this param

itest/opstackl2/op_e2e_test.go
```diff
@@ -40,16 +36,12 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
 
-	// create a mock block
-	r := rand.New(rand.NewSource(time.Now().Unix()))
-	block := &types.BlockInfo{
-		Height: lastCommittedStartHeight,
-		// mock block hash
-		Hash: datagen.GenRandomByteArray(r, 32),
-	}
+	testBlocks := ctm.WaitForNBlocksAndReturn(t, lastCommittedStartHeight, 1)
+	require.Equal(t, 1, len(testBlocks))
```
bap2pecs, line 40: we already hv `require.Equal(t, n, len(blocks))` in the function. no need to check it here again

itest/opstackl2/op_e2e_test.go
```diff
@@ -59,23 +51,49 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
 	)
 	require.NoError(t, err)
 	t.Logf("Submit finality signature to op finality contract")
 
-	// mock more blocks
-	blocks := []*types.BlockInfo{}
+	queryParams := &sdk.L2Block{
+		BlockHeight:    testBlock.Height,
+		BlockHash:      hex.EncodeToString(testBlock.Hash),
+		BlockTimestamp: uint64(1231473952),
```
bap2pecs, line 65: iirc, there were some long doc blocks here explaining about this magic number. do you know if we deleted it by accident?

itest/opstackl2/op_e2e_test.go
```diff
@@ -40,16 +36,12 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
-		// mock block hash
-		Hash: datagen.GenRandomByteArray(r, 32),
-	}
+	testBlocks := ctm.WaitForNBlocksAndReturn(t, lastCommittedStartHeight, 1)
+	require.Equal(t, 1, len(testBlocks))
+	testBlock := testBlocks[0]
 
 	// fp sign
-	fpSig, err := fpInstance.SignFinalitySig(block)
+	fpSig, err := fpInstance.SignFinalitySig(testBlock)
```
bap2pecs, line 44: why not also use `WaitForFpVoteAtHeight`? wouldn't this logic have a potential race condition?

itest/opstackl2/op_e2e_test.go
```diff
@@ -143,25 +161,24 @@ func TestBlockBabylonFinalized(t *testing.T) {
...
-		Height: targetBlockHeight,
-		Hash:   datagen.GenRandomByteArray(r, 32),
-	}
-	ctm.fpSubmitFinalitySignature(t, fpList[0], fpStartHeightList[0], testBlock)
-	ctm.fpSubmitFinalitySignature(t, fpList[1], fpStartHeightList[1], testBlock)
+	ctm.WaitForFpVoteAtHeight(t, fpList[0], targetBlockHeight)
+	// stop the first FP instance
+	fpStopErr := fpList[0].Stop()
+	require.NoError(t, fpStopErr)
+	t.Logf("Stop the first FP instance")
```
bap2pecs, line 176: is it possible that a race condition like this happens?

- wait for vote at height N
- try to stop first FP
- first FP submits vote for height N+1
- first FP finished stopping

how can we prevent it?

itest/opstackl2/op_e2e_test.go
```diff
@@ -189,23 +206,24 @@ func TestBlockBabylonFinalized(t *testing.T) {
...
 
-	// ======  another test case only for the last FP instance ======
-	nextBlockHeight := targetBlockHeight + 1
-	r2 := rand.New(rand.NewSource(time.Now().Unix()))
-	testNextBlock := &types.BlockInfo{
-		Height: nextBlockHeight,
-		Hash:   datagen.GenRandomByteArray(r2, 32),
-	}
-	ctm.fpSubmitFinalitySignature(t, fpList[1], fpStartHeightList[1], testNextBlock)
+	// ======  another test case only for the last FP instance sign ======
```
bap2pecs, line 209: "=== another test case: only the second FP votes; so the block not reach quorum ==="

itest/opstackl2/op_e2e_test.go
```diff
@@ -189,23 +206,24 @@ func TestBlockBabylonFinalized(t *testing.T) {
...
+	// select a suitable height as the next block height:
+	// * the last FP instance had committed the pub rand (e.g. < LastPublicRandCommitHeight)
+	// * not wait too long for the test to finish (e.g. 32, NumPubRand is 64)
+	testNextBlockHeight := *fpStartHeightList[1] + 32
```
bap2pecs, line 213: is the purpose of choosing 32 to prevent the race condition I described above? i.e. 32 is large enough to make sure first FP won't vote the block

itest/opstackl2/op_test_manager.go
```diff
@@ -240,12 +241,42 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
 	t.Logf("Babylon node has started")
 }
 
+func (ctm *OpL2ConsumerTestManager) WaitForNBlocksAndReturn(t *testing.T, startHeight uint64, n int) []*types.BlockInfo {
+	var (
+		blocks []*types.BlockInfo
+		err    error
+	)
```
bap2pecs, line 248: I think it's cleaner to just do

```
var blocks []*types.BlockInfo
var err error
```
to save two lines and one nesting

itest/opstackl2/op_test_manager.go
```diff
@@ -240,12 +241,42 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		blocks, err = ctm.OpL2ConsumerCtrl.QueryBlocks(startHeight, startHeight+uint64(n-1), uint64(n))
+		if err != nil || blocks == nil {
+			return false
+		}
+		return len(blocks) == n
+	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
+	require.Equal(t, n, len(blocks))
+	t.Logf("The last block is %d, %s", blocks[n-1].Height, hex.EncodeToString(blocks[n-1].Hash))
```
bap2pecs, line 257: nit: also print `n`

itest/opstackl2/op_test_manager.go
```diff
@@ -240,12 +241,42 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
...
+func (ctm *OpL2ConsumerTestManager) WaitForFpVoteAtHeight(t *testing.T, fpIns *service.FinalityProviderInstance, height uint64) {
+	lastVotedHeight := fpIns.GetLastVotedHeight()
+	require.Eventually(t, func() bool {
+		if lastVotedHeight >= height {
+			return true
+		} else {
+			lastVotedHeight = fpIns.GetLastVotedHeight()
+			return false
+		}
+	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
```
bap2pecs, line 270: it's cleaner to just do

```
require.Eventually(t, func() bool {
	lastVotedHeight := fpIns.GetLastVotedHeight()
	return lastVotedHeight >= height
}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
```

itest/opstackl2/op_test_manager.go
lesterli: It needs to move them outside of for-loop

itest/opstackl2/op_e2e_test.go
lesterli: To shorten the test time, this param could be deleted

itest/opstackl2/op_test_manager.go
```diff
@@ -240,13 +242,43 @@ func (ctm *OpL2ConsumerTestManager) WaitForServicesStart(t *testing.T) {
...
+	t.Logf("Fp %s voted at height %d", fpIns.GetBtcPkHex(), height)
+}
+
 func (ctm *OpL2ConsumerTestManager) WaitForTargetBlockPubRand(t *testing.T, fpList []*service.FinalityProviderInstance, requiredBlockOverlapLen uint64) []*uint64 {
 	require.Equal(t, 2, len(fpList), "The below algorithm only supports two FPs")
 	fpStartHeightList := make([]*uint64, 2)
-	require.Eventually(t, func() bool {
-		firstFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryLastPublicRandCommit(fpList[0].GetBtcPk())
-		secondFpCommittedPubRand, _ := ctm.OpL2ConsumerCtrl.QueryLastPublicRandCommit(fpList[1].GetBtcPk())
+	firstFpCommittedPubRand, _ := queryFirstPublicRandCommit(ctm.OpL2ConsumerCtrl, fpList[0].GetBtcPk())
```
bap2pecs, line 278: if the error is not needed, we can return form the return value

itest/opstackl2/op_e2e_test.go
lesterli: ok

itest/opstackl2/op_test_manager.go
bap2pecs: simply moving it out is not correct. it's more complicated than that. https://github.com/babylonchain/finality-provider/pull/459/commits/28427dcf71b99b53034bb51ba023cfd6b59ae9a2 fixed it

bap2pecs:
we will fix this in another PR. currently it's using BTC mainnet for some queries so we might hit this rate limit

```
    state.go:371:               ERROR[07-03|02:07:57.463] Derivation process critical error        role=sequencer err="failed to check if block 10 is finalized on Babylon: RPC error: map[code:-31999 data:You’re using more requests than are allowed in your plan. Register to upgrade it for FREE  https://www.ankr.com/rpc/?banner=register message:rate limit exceeded]"
--- FAIL: TestOpMultipleFinalityProviders (91.10s)
FAIL
FAIL    github.com/babylonchain/finality-provider/itest/opstackl2       211.023s
```

itest/opstackl2/op_e2e_test.go
bap2pecs: there is another one `require.Equal(t, 3, len(testBatchBlocks))`. i just removed it

itest/opstackl2/op_e2e_test.go
bap2pecs: i fixed it in https://github.com/babylonchain/finality-provider/pull/459/commits/1327501b16678a40907503ed9533bd7f4b4d58b3

not possible to hv race condition anymore

itest/opstackl2/op_e2e_test.go
bap2pecs: no need to guess a safe number. https://github.com/babylonchain/finality-provider/pull/459/commits/1327501b16678a40907503ed9533bd7f4b4d58b3 fixed it

clientcontroller/opstackl2/msg.go
```diff
@@ -36,8 +36,9 @@ type SubmitFinalitySignatureResponse struct {
 }
 
 type QueryMsg struct {
-	Config            *Config            `json:"config,omitempty"`
-	LastPubRandCommit *LastPubRandCommit `json:"last_pub_rand_commit,omitempty"`
+	Config             *Config             `json:"config,omitempty"`
+	LastPubRandCommit  *LastPubRandCommit  `json:"last_pub_rand_commit,omitempty"`
+	FirstPubRandCommit *FirstPubRandCommit `json:"first_pub_rand_commit,omitempty"`
```
gitferry, line 41: Why do we have them separate structs both containing `BtcPkHex`?

itest/opstackl2/op_e2e_test.go
```diff
@@ -27,8 +43,35 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
...
+
+	// wait for the fp sign
+	ctm.WaitForFpVoteAtHeight(t, fpInstance, testBlock.Height)
+	queryParams := &sdk.L2Block{
+		BlockHeight:    testBlock.Height,
+		BlockHash:      hex.EncodeToString(testBlock.Hash),
+		BlockTimestamp: BlockTimestamp,
+	}
+	_, err = ctm.SdkClient.QueryIsBlockBabylonFinalized(queryParams)
+	require.ErrorIs(t, err, sdk.NoFpHasVotingPowerError)
```
gitferry, line 63: if we expect error here, how do we know the sumission is success?

itest/opstackl2/op_e2e_test.go
```diff
@@ -93,8 +103,8 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	// submit batch finality signatures to smart contract
 	_, err = ctm.OpL2ConsumerCtrl.SubmitBatchFinalitySigs(
```
gitferry, line 104: here we know there's no error of the submission but how do we know the sigs are there? Can we query the contract?

gitferry (approved):
Nice work! Some questions:

clientcontroller/opstackl2/msg.go
bap2pecs: this maps to two CW query msgs: https://github.com/babylonchain/babylon-contract/blob/main/contracts/op-finality-gadget/src/msg.rs#L27-L35

itest/opstackl2/op_e2e_test.go
bap2pecs: `ctm.WaitForFpVoteAtHeight(t, fpInstance, testBlock.Height)` has require.Eventaully to guarantee the submission is successfully at the specified height

```
func (ctm *OpL2ConsumerTestManager) WaitForFpVoteAtHeight(t *testing.T, fpIns *service.FinalityProviderInstance, height uint64) {
	require.Eventually(t, func() bool {
		lastVotedHeight := fpIns.GetLastVotedHeight()
		return lastVotedHeight >= height
	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
	t.Logf("Fp %s voted at height %d", fpIns.GetBtcPkHex(), height)
}
```

itest/opstackl2/op_e2e_test.go
bap2pecs: > Can we query the contract?

created https://github.com/babylonchain/finality-provider/issues/461 as a follow up

clientcontroller/opstackl2/msg.go
maurolacy: You can technically use a single struct here (i.e. `PubRandCommit`). And annotate the JSON query names with `first_`/`last_`.

I would also define the first query on top of the last one. But that's just me. 🙂

lesterli: It makes sense, we will update it with this:
```
type QueryMsg struct {
    Config              *Config        `json:"config,omitempty"`
    FirstPubRandCommit  *PubRandCommit `json:"first_pub_rand_commit,omitempty"`
    LastPubRandCommit   *PubRandCommit `json:"last_pub_rand_commit,omitempty"`
}
```

PR #460, opened 2024-07-02
feat: add export of `ProofOfPossessionBTC` to `eotsd`

RafilxTenfen:
This export of proof of possession is useful for building messages like `MsgCreateFinalityProvider`

Add new command `eotsd pop-export [bbn-addr] ... `

### DIFF
--- a/eotsmanager/cmd/eotsd/daemon/pop.go
+++ b/eotsmanager/cmd/eotsd/daemon/pop.go
@@ -1,0 +1,141 @@
+package daemon
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/cometbft/cometbft/crypto/tmhash"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	bbn "github.com/babylonchain/babylon/types"
+	btcstktypes "github.com/babylonchain/babylon/x/btcstaking/types"
+	"github.com/babylonchain/finality-provider/eotsmanager"
+	"github.com/babylonchain/finality-provider/eotsmanager/config"
+	"github.com/babylonchain/finality-provider/log"
+	"github.com/urfave/cli"
+)
+
+// PoPExport the data for exporting the PoP.
+// The PubKeyHex is the public key of the finality provider BTC key to load
+// the private key and sign the AddressSiged.
+type PoPExport struct {
+	PubKeyHex     string                           `json:"pub_key_hex"`
+	PoP           btcstktypes.ProofOfPossessionBTC `json:"pop"`
+	PoPHex        string                           `json:"pop_hex"`
+	AddressSigned string                           `json:"address_signed"`
+}
+
+var ExportPoPCommand = cli.Command{
+	Name:      "pop-export",
+	Usage:     "Exports the Proof of Possession by Signing over the finality provider address with the EOTS private key.",
+	UsageText: "pop-export [bbn-address]",
+	Description: `Parse the address received as argument, hash it with
+	sha256 and sign based on the Schnorr key associated with the key-name or btc-pk flag.
+	If the both flags are supplied, btc-pk takes priority. Use the generated signature
+	to build a Proof of Possession and exports it.`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  homeFlag,
+			Usage: "Path to the keyring directory",
+			Value: config.DefaultEOTSDir,
+		},
+		cli.StringFlag{
+			Name:  keyNameFlag,
+			Usage: "The name of the key to load private key for signing",
+		},
+		cli.StringFlag{
+			Name:  fpPkFlag,
+			Usage: "The public key of the finality-provider to load private key for signing",
+		},
+		cli.StringFlag{
+			Name:  passphraseFlag,
+			Usage: "The passphrase used to decrypt the keyring",
+			Value: defaultPassphrase,
+		},
+		cli.StringFlag{
+			Name:  keyringBackendFlag,
+			Usage: "The backend of the keyring",
+			Value: defaultKeyringBackend,
+		},
+	},
+	Action: ExportPoP,
+}
+
+func ExportPoP(ctx *cli.Context) error {
+	keyName := ctx.String(keyNameFlag)
+	fpPkStr := ctx.String(fpPkFlag)
+	passphrase := ctx.String(passphraseFlag)
+	keyringBackend := ctx.String(keyringBackendFlag)
+
+	args := ctx.Args()
+	bbnAddressStr := args.First()
+	if len(bbnAddressStr) == 0 {
+		return errors.New("invalid argument, please provide a valid bbn address as argument")
+	}
+
+	bbnAddr, err := sdk.AccAddressFromBech32(bbnAddressStr)
+	if err != nil {
+		return fmt.Errorf("invalid argument %s, please provide a valid bbn address as argument, err: %w", bbnAddressStr, err)
+	}
+
+	if len(fpPkStr) == 0 && len(keyName) == 0 {
+		return fmt.Errorf("at least one of the flags: %s, %s needs to be informed", keyNameFlag, fpPkFlag)
+	}
+
+	homePath, err := getHomeFlag(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to load home flag: %w", err)
+	}
+
+	cfg, err := config.LoadConfig(homePath)
+	if err != nil {
+		return fmt.Errorf("failed to load config at %s: %w", homePath, err)
+	}
+
+	logger, err := log.NewRootLoggerWithFile(config.LogFile(homePath), cfg.LogLevel)
+	if err != nil {
+		return fmt.Errorf("failed to load the logger")
+	}
+
+	dbBackend, err := cfg.DatabaseConfig.GetDbBackend()
+	if err != nil {
+		return fmt.Errorf("failed to create db backend: %w", err)
+	}
+	defer dbBackend.Close()
+
+	eotsManager, err := eotsmanager.NewLocalEOTSManager(homePath, keyringBackend, dbBackend, logger)
+	if err != nil {
+		return fmt.Errorf("failed to create EOTS manager: %w", err)
+	}
+
+	hashOfMsgToSign := tmhash.Sum(bbnAddr.Bytes())
+	btcSig, pubKey, err := singMsg(eotsManager, keyName, fpPkStr, passphrase, hashOfMsgToSign)
+	if err != nil {
+		return fmt.Errorf("failed to sign address %s: %w", bbnAddr.String(), err)
+	}
+
+	bip340Sig := bbn.NewBIP340SignatureFromBTCSig(btcSig)
+	btcSigBz, err := bip340Sig.Marshal()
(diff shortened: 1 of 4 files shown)


eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	PoPHex        string                           `json:"pop_hex"`
+	AddressSigned string                           `json:"address_signed"`
+}
+
+var ExportPoPCommand = cli.Command{
+	Name:      "pop-export",
+	Usage:     "Exports the Proof of Possession by Signing over the finality provider address with the EOTS private key.",
+	UsageText: "pop-export [bbn-address]",
+	Description: `Parse the address received as argument, hash it with
+	sha256 and sign based on the Schnorr key associated with the key-name or btc-pk flag.
```
gitferry, line 33:
```suggestion
	sha256 and sign based on the EOTS key associated with the key-name or btc-pk flag.
```

I think we use Schnorr/BTC/EOTS/FinalityProvider key interchangeably, which might be confusing for users. Maybe we just say EOTS key here

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	If the both flags are supplied, btc-pk takes priority. Use the generated signature
+	to build a Proof of Possession and exports it.`,
```
gitferry, line 35:
```suggestion
	to build a Proof of Possession and export it.`,
```

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	keyName := ctx.String(keyNameFlag)
+	fpPkStr := ctx.String(fpPkFlag)
+	passphrase := ctx.String(passphraseFlag)
+	keyringBackend := ctx.String(keyringBackendFlag)
+
+	args := ctx.Args()
+	bbnAddressStr := args.First()
+	if len(bbnAddressStr) == 0 {
+		return errors.New("invalid argument, please provide a valid bbn address as argument")
+	}
```
gitferry, line 74: I think this case is already covered by `sdk.AccAddressFromBech32` below

eotsmanager/cmd/eotsd/daemon/pop_test.go
```diff
@@ -0,0 +1,67 @@
...
+		bbnAddr := datagen.GenRandomAccount().GetAddress()
+
+		btcPkFlag := fmt.Sprintf("--btc-pk=%s", keyOut.PubKeyHex)
+		exportedPoP := appRunPoPExport(r, t, app, []string{bbnAddr.String(), hFlag, btcPkFlag})
+		require.NotNil(t, exportedPoP)
+		require.NoError(t, exportedPoP.PoP.ValidateBasic())
+
+		btcPubKey, err := bbn.NewBIP340PubKeyFromHex(exportedPoP.PubKeyHex)
+		require.NoError(t, err)
+		require.NoError(t, exportedPoP.PoP.Verify(bbnAddr, btcPubKey, &chaincfg.MainNetParams))
```
gitferry, line 52:
```suggestion
		require.NoError(t, exportedPoP.PoP.Verify(bbnAddr, btcPubKey, &chaincfg.SigNetParams))
```

We use `signet` in default but here btc net params does not matter. Sill better to keep it consistent

gitferry (approved):
LGTM! Minor comments:

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/babylonchain/finality-provider/eotsmanager"
+	"github.com/babylonchain/finality-provider/eotsmanager/config"
+	"github.com/babylonchain/finality-provider/log"
+	"github.com/urfave/cli"
+)
+
+// PoPExport the data for exporting the PoP.
+// The PubKeyHex is the public key of the finality provider BTC key to load
+// the private key and sign the AddressSiged.
+type PoPExport struct {
```
SebastianElvis, line 21: I'm not sure if this command is expressive enough. What's the use case for it and will it be covering more use cases if we export the entire finality provider struct?

RafilxTenfen: My idea here was to be able to export the PoP to build the signed msg `MsgCreateFinalityProvider` using `babylond tx btcstaking create-finality-provider --offline`

RafilxTenfen: on that command, babylon expects the pop ready as hex on the CLI

eotsmanager/cmd/eotsd/daemon/pop_test.go
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/460/commits/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/460/commits/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/460/commits/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	PubKeyHex     string                           `json:"pub_key_hex"`
+	PoP           btcstktypes.ProofOfPossessionBTC `json:"pop"`
+	PoPHex        string                           `json:"pop_hex"`
+	AddressSigned string                           `json:"address_signed"`
+}
+
+var ExportPoPCommand = cli.Command{
+	Name:      "pop-export",
+	Usage:     "Exports the Proof of Possession by Signing over the finality provider address with the EOTS private key.",
```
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/460/commits/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/urfave/cli"
+)
+
+// PoPExport the data for exporting the PoP.
+// The PubKeyHex is the public key of the finality provider BTC key to load
+// the private key and sign the AddressSiged.
+type PoPExport struct {
+	PubKeyHex     string                           `json:"pub_key_hex"`
+	PoP           btcstktypes.ProofOfPossessionBTC `json:"pop"`
+	PoPHex        string                           `json:"pop_hex"`
```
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/460/commits/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
RafilxTenfen: https://github.com/babylonchain/finality-provider/commit/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,141 @@
...
+	AddressSigned string                           `json:"address_signed"`
```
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonchain/finality-provider/commit/d1c66761d2fbe7d9ca9ab7ddb472ccb3db28cc26

eotsmanager/cmd/eotsd/daemon/pop.go
```diff
@@ -0,0 +1,139 @@
...
+	dbBackend, err := cfg.DatabaseConfig.GetDbBackend()
+	if err != nil {
+		return fmt.Errorf("failed to create db backend: %w", err)
+	}
+	defer dbBackend.Close()
+
+	eotsManager, err := eotsmanager.NewLocalEOTSManager(homePath, keyringBackend, dbBackend, logger)
+	if err != nil {
+		return fmt.Errorf("failed to create EOTS manager: %w", err)
+	}
```
SebastianElvis, line 108: Is it possible that the finality provider worked with a remote EOTS signer and the local DB does not preserve the BTC key in FP?

eotsmanager/cmd/eotsd/daemon/pop.go
SebastianElvis: I see. Since the end goal is to build a `MsgCreateFinalityProvider` message, how about we make this command to export `MsgCreateFinalityProvider` message directly? This includes everything and might streamline the UX as users no longer need to manually wire up `MsgCreateFinalityProvider`

SebastianElvis (approved):

eotsmanager/cmd/eotsd/daemon/pop.go
gitferry: This is a good point. The fpd program is responsible for sending `MsgCreateFinalityProvider` to Babylon for registration anyway. So I would imagine there would be a command to construct `MsgCreateFinalityProvider` in an off-line mode

RafilxTenfen: We could, but `babylond` already has this command implemented in `babylond tx btcstaking create-finality-provider --generate-only` and latter `babylond tx sign` to sign the generated msg

eotsmanager/cmd/eotsd/daemon/pop.go
RafilxTenfen: I did not thought of this possibility, this is the same used for the schnorr sign command, if this is required it might need some refactory

eotsmanager/cmd/eotsd/daemon/pop.go
RafilxTenfen: It would make life easier for finality providers, but the only reason this is not done in the babylond is because at the chain node should not load BTC private key to create the PoP

gitferry: So, finality providers would need to execute two commands using two different programs to generate the registration message? It would be great to finish everthing using `fpd` as it has to generate btc sig anyway

RafilxTenfen: 3 commands or more, since they need to create keys...
But a reminder that they should be technical people, they will need to have some expertise to run the finality provider daemon when it is needed

gitferry: I understand but if there's a chance where less commands are needed, we should do it. Otherwise, it would be overhead for us to maintain the doc and do Q&A

PR #462, opened 2024-07-03
Fix/base64 encoding

maurolacy:
Cleans the base64 mess up by reverting relevant fields to `[]byte`, and removing explicit Base64 encoding.

There may be more cases to fix / clean up around. Particularly in the op controller.

Also, @gusin13, though having explicit structs for mapping is generally a good idea, most of them might not be needed now, and could be removed (i.e. directly using the pb struct when possible / convenient).


clientcontroller/opstackl2/consumer.go
```diff
@@ -155,7 +154,7 @@ func (cc *OPStackL2ConsumerController) SubmitFinalitySig(
 			FpPubkeyHex: bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex(),
 			Height:      block.Height,
 			PubRand:     bbntypes.NewSchnorrPubRandFromFieldVal(pubRand).MustMarshal(),
-			Proof:       ConvertProof(cmtProof),
+			Proof:       cmtProof,
```
bap2pecs, line 157: this implies we changed the data passed to the CW contract. do we need to modify anything in the CW contract?

maurolacy: No. This works just fine, because on the contract side the byte array fields are marked as `Binary`.

To sum up: pb structs with JSON support can be used just fine on the client (Go) side. On the contract side, unfortunately they require the `Binary` wrapper for byte array fields to decode correctly.

bap2pecs: got it. that's good to know

bap2pecs (commented):
nice fix! OP parts lgtm

bap2pecs:
@maurolacy quick tip, to test op locally, you need to

1. download devnet data
2. get your own BTC mainet RPC and replace it to avoid rate limiting issue (you can get one from quicknode)

see instructions https://github.com/babylonchain/finality-provider/tree/base/consumer-chain-support/itest/opstackl2

maurolacy:
I see, running into that now. Can you take a look at the failing OP test btw? If not, I can do it tomorrow.

bap2pecs:
sure I am looking at the OP test now.

clientcontroller/cosmwasm/msg.go
```diff
@@ -140,24 +142,17 @@ type CommitPublicRandomness struct {
...
-	Height      uint64 `json:"height"`
-	PubRand     string `json:"pub_rand"`   // base64 encoded
-	Proof       Proof  `json:"proof"`      // nested struct
-	BlockHash   string `json:"block_hash"` // base64 encoded
-	Signature   string `json:"signature"`  // base64 encoded
-}
-
-type Proof struct {
-	Total    uint64   `json:"total"`
-	Index    uint64   `json:"index"`
```
bap2pecs, line 158: @maurolacy if you change the logger level in `StartOpL2ConsumerManager` to `zapcore.DebugLeve`,

you can see
```
2024-07-03T17:20:04.562-0400	DEBUG	service/fp_instance.go:567	failed to submit finality signature to the consumer chain	{"pk": "a505229673b9385c745eee345e4970bc97bee97ae02018bbc6f6837bdadcd1ba", "current_failures": 3, "target_block_height": 1, "error": "failed to send finality signature to the consumer chain: rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: Error parsing into type op_finality_gadget::msg::ExecuteMsg: missing field `index`: execute wasm contract failed [CosmWasm/wasmd@v0.51.0/x/wasm/keeper/keeper.go:422] with gas used: '103566': unknown request"}
```

I think removing this "index" field is the culprit of the failure

bap2pecs:
@maurolacy btw you can merge https://github.com/babylonchain/finality-provider/pull/466 here. this fixed some flakiness in our op tests

maurolacy:
> @maurolacy quick tip, to test op locally, you need to
> 1. download devnet data
> 2. get your own BTC mainet RPC and replace it to avoid rate limiting issue (you can get one from quicknode)
> 
> see instructions https://github.com/babylonchain/finality-provider/tree/base/consumer-chain-support/itest/opstackl2

Tests should be self-contained. Ins't there a script or pre-processing step we can use to automate this?

clientcontroller/cosmwasm/msg.go
maurolacy: Index is there in the other struct as well. The only change is from `uint64` to `int64`.

maurolacy:
Reverted the OP changes and now tests pass.

I suggest OP guys (@lesterli, @bap2pecs) check that reverted commit and incorporate the changes there if warranted.

bap2pecs:
> Tests should be self-contained. Ins't there a script or pre-processing step we can use to automate this?

yes there is a command in the README https://github.com/babylonchain/finality-provider/tree/base/consumer-chain-support/itest/opstackl2

```
$ make op-e2e-devnet
```

this is already automated in CI. but since the test data is in gitignore, you still need to run it locally, just once

bap2pecs:
> Reverted the OP changes and now tests pass.
> 
> I suggest OP guys (@lesterli, @bap2pecs) check that reverted commit and incorporate the changes there if warranted.

happy to that later. but I am just worried that we are not fixing the root cause. if we add the same tests in the other CosmWasm controller, it might fail again (my guess is we are missing that test there)

the error message shows

```
Error parsing into type op_finality_gadget::msg::ExecuteMsg: missing field `index`: execute wasm contract failed
```

which seems relevant to this PR

SebastianElvis (approved):
Lgtm! nice finding

maurolacy:
Fixed the OP tests but now the `build_docker` test target is failing:

https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1757/workflows/5c9b4489-8212-4d51-a012-a95d826a1e95/jobs/4454

Does anybody know what may be happening?

go.mod
```diff
@@ -431,7 +431,7 @@ require (
 replace (
 	// use cosmos fork of keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
-	github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.8.6-0.20240629034230-f5dbfa1d9875
+	github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.8.6-0.20240620002950-c5a8d317091e
```
bap2pecs, line 434: this is why the CI fails 

```
panic: proto: could not resolve import "core/crypto/pb/crypto.proto": not found [recovered]
        panic: proto: could not resolve import "core/crypto/pb/crypto.proto": not found
```

same thing @SebastianElvis and we were discussing in the group

can we change this back for now?

maurolacy: Unfortunately we cannot, as the original ref (`f5dbfa1d9875 `) has disappeared / is not found.

I suggest we disable op e2e tests in CI for now, until this issue is solved one way or another.

bap2pecs: ok I am adding a FIXME there and will update this PR soon

bap2pecs:
all CI passed. merging

Issue #465, opened 2024-07-03
consolidate LastPubRandCommit and FirstPubRandCommit


bap2pecs:
https://github.com/babylonchain/finality-provider/pull/459#discussion_r1664061465

PR #466, opened 2024-07-03
chore: op query pub rand with `PubRandCommit`

lesterli:
This PR does these two updates:
* Close https://github.com/babylonchain/finality-provider/issues/465, replaces `FirstPubRandCommit` and `LastPubRandCommit` with `PubRandCommit`.
*  Use the latest commit for `babylon-da-sdk` and `babylonchain/optimism`


bap2pecs:
CI failed b/c we are using the mainnet BTC rpc which can get rate limited. so we are thinking of setting up a local btc network for e2e test. @gitferry I see you have done https://github.com/babylonchain/staking-indexer/pull/5/files before. do you think we can do the same set up here?

gusin13 (approved):

bap2pecs:
found another flaky test

```
=== RUN   TestMultipleFinalityProviders
    babylon_node_handler.go:148: babylon log file: /tmp/zBabylonTest1947099691/babylon.log
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
    babylon_test_manager.go:113: Babylon node is started
    babylon_test_manager.go:169: The test manager is running with 3 finality-provider(s)
fatal error: concurrent map writes
fatal error: concurrent map writes
```

PR #472, opened 2024-07-05
chore: add debug logs for op consumer controller

lesterli:
This PR adds the debug logs for the OP consumer controller.

```
make lint
```

### DIFF
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -235,13 +235,11 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
 	if err != nil {
 		return nil, err
 	}
-	for _, block := range blocks {
-		cc.logger.Debug(
-			"Successfully submitted finality signatures in a batch",
-			zap.Uint64("height", block.Height),
-			zap.String("block_hash", hex.EncodeToString(block.Hash)),
-		)
-	}
+	cc.logger.Debug(
+		"Successfully submitted finality signatures in a batch",
+		zap.Uint64("start_height", blocks[0].Height),
+		zap.Uint64("end_height", blocks[len(blocks)-1].Height),
+	)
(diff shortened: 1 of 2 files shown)


clientcontroller/opstackl2/consumer.go
```diff
@@ -182,6 +183,11 @@ func (cc *OPStackL2ConsumerController) SubmitFinalitySig(
 	if err != nil {
 		return nil, err
 	}
+	cc.logger.Debug(
+		"Submit finality signature",
```
gitferry, line 187:
```suggestion
		"Successfully submitted finality signature",
```

clientcontroller/opstackl2/consumer.go
```diff
@@ -229,7 +235,13 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
...
 		return nil, err
 	}
-
+	for _, block := range blocks {
+		cc.logger.Debug(
+			"Submit batch finality signatures",
+			zap.Uint64("height", block.Height),
+			zap.String("block_hash", hex.EncodeToString(block.Hash)),
+		)
+	}
```
gitferry, line 244: Is it neceesary to iterate all the blocks? Can we do like this:

```
cc.logger.Debug(
	"Successfully submitted finality signatures in a batch",
		zap.Uint64("start_height", blocks[0].Height),
                zap.Uint64("end_height", blocks[len(blocks)-1].Height),
	)
```

gitferry (approved):

clientcontroller/opstackl2/consumer.go
lesterli: When debugging today, we needed to check the block hash of fp signs, and thus discovered that the bug is caused by inconsistent hash values. https://github.com/babylonchain/babylon-da-sdk/pull/44/files

gitferry: OK then. Let's have the loop

PR #474, opened 2024-07-05
bump: both fp and `tools` use the same Babylon version

lesterli:
This PR updates the deps in the `go.mod` to ensure finality-provider and the `tools` use [the same version of Babylon](https://github.com/babylonchain/babylon-private/commit/c5a8d317091e2965e20ea56fa10e98d34aaa3547).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -431,7 +431,7 @@ require (
 replace (
 	// use cosmos fork of keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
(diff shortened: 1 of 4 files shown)


SebastianElvis (approved):

bap2pecs:
@SebastianElvis this actually still has the issue we encountered earlier. so making the version the same won't fix the problem

bap2pecs:
no need to have this anymore since the tmp fix is merged https://github.com/babylonchain/babylon-private/pull/264

we created https://github.com/babylonchain/babylon-private/issues/266 to track the issue and currently asking the author that introduced the code in https://github.com/CosmWasm/wasmd/issues/1785

cc @SebastianElvis

Issue #473, opened 2024-07-05
properly json unmarshal


bap2pecs:
the right way to handle "null" in json is to use pointer type

example:

```
	if resp.Data == nil || string(resp.Data) == "null" {
		return nil, nil
	}
	var votedFpPkHexList []string
	if err := json.Unmarshal(resp.Data, &votedFpPkHexList); err != nil {
		return nil, err
	}
```

=> 

```
	votedFpPkHexList := &[]string{}
	if err := json.Unmarshal(resp.Data, votedFpPkHexList); err != nil {
		return nil, err
	}
```

there are a few places in this repo that has ^ pattern. let'x fix

PR #443, opened 2024-06-28
test: add test case for op stack e2e test

lesterli:
## Summary

This PR adds a test case for the op stack e2e test.

## Test Plan

```
make test
make test-e2e
```

PR #483, opened 2024-07-07
chore: 1) fix null JSON resp handling; 2) op e2e test clean-up 

lesterli:
This PR fixes a few issues:

- [x] proper handling of JSON response with "null" (#473)
- [x] e2e test cleanup:
	- remove unused `BlockTimestamp` (context, after https://github.com/babylonchain/babylon-da-sdk/pull/45, we have decoupled e2e test w mainnet BTC RPC)
	- clean up `StartFinalityProvider` (since we already have `RegisterBBNFinalityProvider`)
	- removed unused fields in `OpL2ConsumerTestManager`
	- added `getConsumerChainId()` so we get rid of the `OpChainId` field 
	- start EOTS manager earlier (might be useful to fix the flaky "the EOTS manager server is not responding" issue
	- address https://github.com/babylonchain/finality-provider/pull/443#discussion_r1660615148 by setting cw contract address in the config before calling `NewOPStackL2ConsumerController` so we don't need to generate a valid random address first to bypass validation check


itest/opstackl2/op_e2e_test.go
```diff
@@ -57,7 +38,7 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	queryParams := &sdk.L2Block{
 		BlockHeight:    testBlock.Height,
 		BlockHash:      hex.EncodeToString(testBlock.Hash),
-		BlockTimestamp: BlockTimestamp,
+		BlockTimestamp: 12345, // doesn't matter b/c the BTC client is mocked
```
bap2pecs, line 41: context, after https://github.com/babylonchain/babylon-da-sdk/pull/45, we have decoupled e2e test w mainnet BTC RPC

SebastianElvis: Btw, besides mocking BTC node, one can also leverage the https://github.com/btcsuite/btcd/tree/master/integration/rpctest package for testing things against an almost real BTC node. But this can be done in the future.

SebastianElvis (approved):

PR #485, opened 2024-07-07
test: add test util to prefix t.Logf with date string

bap2pecs:
## Summary

I found this helpful at debugging. after the change, we will see sth like

```
log.go:12: [14:15:52.270] Register consumer op-stack-l2-901 to Babylon
```

## Test Plan

```
make lint
make test-e2e-op
```


itest/opstackl2/op_e2e_test.go
```diff
@@ -48,7 +49,7 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	committedPubRand, err := queryFirstPublicRandCommit(ctm.OpL2ConsumerCtrl, fpInstance.GetBtcPk())
 	require.NoError(t, err)
 	committedStartHeight := committedPubRand.StartHeight
-	t.Logf("First committed pubrandList startHeight %d", committedStartHeight)
+	log.Logf(t, "First committed pubrandList startHeight %d", committedStartHeight)
```
SebastianElvis, line 52: Can this be done by having a wrapper function for `t.Logf` rather than introducing a new logger?

SebastianElvis: Is there a more native way for doing it, e.g., modifying log format under `t` object?

bap2pecs: I did some research but couldn't find it

SebastianElvis: Did some research and it seems so :sweat:

SebastianElvis (approved):

PR #489, opened 2024-07-08
cosmwasm: fix check of activated height in CosmWasm controller

SebastianElvis:
This PR fixes a bug in CosmWasm consumer controller where we need to check whether the unmarshalled activated height is 0 or not. If it's 0 then it means we are unmarshalling something fishy and actually BTC staking is not activated, and we need to return an error, as [expected](https://github.com/babylonchain/finality-provider/blob/fee9b10e886a098ebd97b7ccfcf9e31bf465a746/finality-provider/service/chain_poller.go#L211-L220) by RPC poller. This fix allows us to remove some hacks and makes e2e / local deployment pass.

@gitferry Do you think in the [poller](https://github.com/babylonchain/finality-provider/blob/a531ebd36686383eb1424929ff95ca820a9b3c5c/finality-provider/service/chain_poller.go#L213-L214) we need to distinguish between the "BTC staking is not activated" and other errors? This seems necessary as it's possible that the errors are some other serious ones, rather than that BTC staking is not activated.

### DIFF
--- a/clientcontroller/cosmwasm/consumer.go
+++ b/clientcontroller/cosmwasm/consumer.go
@@ -246,32 +246,17 @@ func (wc *CosmwasmConsumerController) QueryFinalityProvidersByPower() (*Consumer
 }
 
 func (wc *CosmwasmConsumerController) QueryLatestFinalizedBlock() (*fptypes.BlockInfo, error) {
-	//isFinalized := true
-	//limit := uint64(1)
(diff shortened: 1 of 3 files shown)


maurolacy (approved):
LGTM.

gitferry:
@SebastianElvis sorry for the delay, yeah, separating the error is a good idea!

PR #490, opened 2024-07-08
chore: fpd to use cobra CLI

RafilxTenfen:
Update to use `"github.com/spf13/cobra"` as CLI command in `fpd`
Add `keys` subcommand imported from cosmos sdk


finality-provider/cmd/fpd/main.go
```diff
@@ -4,23 +4,64 @@ import (
...
+		Use:           "fpd",
+		Short:         "fpd - Finality Provider Daemon (fpd).",
+		Long:          `fpd is the daemon to handle finality provider submission of randomness from BTC to proof of stake chains`,
+		SilenceErrors: false,
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			// set the default command outputs
+			cmd.SetOut(cmd.OutOrStdout())
+			cmd.SetErr(cmd.ErrOrStderr())
+
+			clientCtx = clientCtx.WithCmdContext(cmd.Context()).WithViper("")
```
gitferry, line 35: This seems to read client config which we don't have. Maybe we don't need it?

finality-provider/cmd/fpd/main.go
```diff
@@ -4,23 +4,64 @@ import (
...
 
-func fatal(err error) {
-	fmt.Fprintf(os.Stderr, "[btc-finality-provider] %v\n", err)
-	os.Exit(1)
+// NewRootCmd creates a new root command for fpd. It is called once in the main function.
+func NewRootCmd() *cobra.Command {
+	var (
+		clientCtx client.Context
+	)
+	encCfg := appparams.DefaultEncodingConfig()
```
gitferry, line 23: I think this allows us to create account with prefix `bbn`. What if we want to add keys for a consumer chain that does not have such prefix?

finality-provider/cmd/fpd/main.go
```diff
@@ -4,23 +4,64 @@ import (
...
+
+	rootCmd := &cobra.Command{
+		Use:           "fpd",
+		Short:         "fpd - Finality Provider Daemon (fpd).",
+		Long:          `fpd is the daemon to handle finality provider submission of randomness from BTC to proof of stake chains`,
```
gitferry, line 28:
```suggestion
		Long:          `fpd is the daemon to create and manage finality providers`,
```
We can make it general because it also handles submission of finality signatures and serving PoS chains beyond (e.g., L2)

finality-provider/cmd/fpd/main.go
```diff
@@ -4,23 +4,64 @@ import (
...
+			clientCtx.Codec = encCfg.Codec
+			clientCtx, err := client.ReadPersistentCommandFlags(clientCtx, cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			clientCtx, err = config.ReadFromClientConfig(clientCtx)
+			if err != nil {
+				return err
+			}
```
gitferry, line 45: same here

gitferry (commented):
I think it's nice to have cosmos commands for free by importing cobra CLI. Some concerns:
1. the `keys` command seems to only generate account for `bbn`. How can we customize it to work with other consumers?
2. the `keys` command needs extra config file `config/client.toml` and it cannot change the `Key` field of `fpd.conf`. How can we workaround?

SebastianElvis:
I remember @KonradStaniec deliberately avoids cobra and mentioned some drawbacks of it. Maybe Konrad can shed some light on this?

finality-provider/cmd/fpd/main.go
RafilxTenfen: it is used later for setting the `homedir` into the context (which is being used as default home dir of `keys` subcommand)

finality-provider/cmd/fpd/main.go
RafilxTenfen: yeap, we might need to address this when needed, but currently `dev` branch also does not ask for the prefix 

We could have a list of encoding configs and use the one set in a flag like asking for `--bech-prefix`

finality-provider/cmd/fpd/main.go
RafilxTenfen: removed https://github.com/babylonchain/finality-provider/pull/490/commits/7c30a568bae7a4ff183fd4877040d843dc14ba80

finality-provider/cmd/fpd/main.go
RafilxTenfen: oh okay, the viper part
understood, removed

RafilxTenfen:
> I think it's nice to have cosmos commands for free by importing cobra CLI. Some concerns:
> 
> 1. the `keys` command seems to only generate account for `bbn`. How can we customize it to work with other consumers?
> 2. the `keys` command needs extra config file `config/client.toml` and it cannot change the `Key` field of `fpd.conf`. How can we workaround?

For 1. we might have a list of supported prefixes to choose with a flag and then use the correspondent encoding config
for 2. I used the postrun to update the config like it was doing before https://github.com/babylonchain/finality-provider/pull/490/commits/769ed18ed3e52945ae811b53f7251bd687b86614
Let me know what you think @gitferry

KonradStaniec:
>I remember @KonradStaniec deliberately avoids cobra and mentioned some drawbacks of it. Maybe Konrad can shed some light on this?

There are no specific drawbacks, it just that cobra is big library with broad api. For our cli programs we need to use only fraction of that api and cobra is just overkill.

The https://github.com/urfave/cli is smaller and more focused library, which is enough for our use cases.

Ultimately it comes down to personal preference.

Though I must say having option to bring commands from Babylon, is nice `+` for Cobra. So if you guys Cobra makes sense due to this reason, lets go with it 👍

gitferry (approved):
Great work! Thanks for addressing my concerns!

PR #484, opened 2024-07-07
test: fix and re-enable fast sync

lesterli:
This PR enables the fast sync for the op e2e test.

* remove one test case that is tested already by fast sync
* replace the `ethclient.DialContext` with the `ethclient.Dial` to fix the fast sync error: 

```
unexpected fault address 0x13698d21c
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0x13698d21c pc=0x102b6d854]
```

<details>
<summary>Error detail trace</summary>

```
make test-e2e-op
No processes to kill
rm -rf ~/.babylond ~/.wasmd ~/.bcd
cd tools; \
	go install -trimpath github.com/babylonchain/babylon/cmd/babylond
=== RUN   TestOpMultipleFinalityProviders
    babylon_node_handler.go:148: babylon log file: /var/folders/0w/6nn8jssj0nq6jr63qhm8s1hh0000gn/T/zBabylonTest2354903250/babylon.log
Starting babylond with command:  /Users/lyy/go/bin/babylond start --home=/var/folders/0w/6nn8jssj0nq6jr63qhm8s1hh0000gn/T/zBabylonTest2354903250/node0/babylond --log_level=trace --trace
    op_test_manager.go:86: Fp home dir: /var/folders/0w/6nn8jssj0nq6jr63qhm8s1hh0000gn/T/fpe2etest1522777673/fp-home
2024-07-07T19:43:47.797-0400	INFO	cosmos/log.go:131	Successful transaction	{"chain_id": "chain-test", "gas_used": 2690989, "fees": "8016ubbn", "fee_payer": "\ufffdq\ufffdl\ufffd{\ufffdd\ufffd\u0004\ufffd\ufffdO\ufffd\u0007\ufffd\ufffd\ufffd\u001f\ufffd", "height": 2, "msg_types": ["/cosmwasm.wasm.v1.MsgStoreCode"], "tx_hash": "7D40FAF061758712C7FF93322CF75D2F4E2EC7CB5D04E157E2B7098C8A486661"}
[...]

PR #492, opened 2024-07-08
chore: remove hacks in cosmwasm consumer client

gusin13:


clientcontroller/cosmwasm/consumer.go
```diff
@@ -590,6 +572,7 @@ func (wc *CosmwasmConsumerController) queryCometBestBlock() (*fptypes.BlockInfo,
 	}, nil
 }
 
+//nolint:unused
```
bap2pecs, line 575: does lint complain b/c of build tags?

itest/cosmwasm/bcd/bcd_test_manager.go
```diff
@@ -83,6 +83,8 @@ func StartBcdTestManager(t *testing.T) *BcdTestManager {
 	bc, err := bbncc.NewBabylonController(cfg.BabylonConfig, &cfg.BTCNetParams, logger)
 	require.NoError(t, err)
 
+	cfg.FastSyncInterval = 0 // disable fast sync
```
bap2pecs, line 86: why this is disabled? did you meet w any issues as we did in https://github.com/babylonchain/finality-provider/pull/484?

bap2pecs: i see. feel free to check if there is anything you can reuse from our tests. iirc, we have some helper functions to wait for delegation to be active

SebastianElvis (approved):
Lgtm, great work on getting rid of all hacks!

SebastianElvis:
I realised a strong argument that we should stick to the hacks in the two queries. It's possible that the consuemr system only wants to integrate with 0.3.0 that has finality signature submissions but not 0.4.0 that has on-chain finality round, as  0.4.0 needs breaking chanegs in their code base. If we query indexed blocks in the contract then the FP won't work with 0.3.0. If we query Comet blocks then there is no such problem

maurolacy (approved):
LGTM.

Will try this in the deployment repo next.

maurolacy:
> It's possible that the consuemr system only wants to integrate with 0.3.0 that has finality signature submissions but not 0.4.0 that has on-chain finality round, as 0.4.0 needs breaking chanegs in their code base.

I would say, they should attempt those changes, sooner that later. I would go for the 'latest and greatest' 0.4.0 directly.

maurolacy:
After offline talks, we can release this in time (after it works) as a 0.4.1 release.

SebastianElvis:
As per yesterday's discussion, probably let's simply query Cosmos SDK blocks rather than indexed blocks such that FP remains compatible with chains with or without Babylon SDK?

maurolacy:
Closing without merging, after discussion.

maurolacy:
Feel free to re-open if some changes here are to be merged.

PR #493, opened 2024-07-08
chore: migrate fpcli to cobra CLI

RafilxTenfen:
- move `fpcli` subcommands to `fpd`
- use `cobra` instead of `urfave`
- update docs 
- Branched out from #490 (should be merged after)


finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -0,0 +1,393 @@
...
+	jsonBytes, err := json.MarshalIndent(resp, "", "    ")
+	if err != nil {
+		fmt.Println("unable to decode response: ", err)
+		return
+	}
+
+	fmt.Printf("%s\n", jsonBytes)
+}
+
+func loadKeyName(ctx client.Context, cmd *cobra.Command) (string, error) {
```
gitferry, line 371:
```suggestion
func loadKeyName(homeDir string;, cmd *cobra.Command) (string, error) {
```

Seems we can just pass home dir here?

finality-provider/cmd/cmd.go
```diff
@@ -0,0 +1,52 @@
...
+) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		clientCtx, err := client.GetClientQueryContext(cmd)
+		if err != nil {
+			return err
+		}
+
+		return fRunWithCtx(clientCtx, cmd, args)
+	}
+}
```
gitferry, line 52: It seems a bit over-kill to have the two functions here as I can see all of our commands only need `--home` from the `ctx`. Is there another way to have `--home` as global flag?

gitferry (commented):
Big fan of the migration! Some comments:

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -0,0 +1,393 @@
...
+	f.String(monikerFlag, "", "A human-readable name for the finality provider")
+	f.String(identityFlag, "", "An optional identity signature (ex. UPort or Keybase)")
+	f.String(websiteFlag, "", "An optional website link")
+	f.String(securityContactFlag, "", "An email for security contact")
+	f.String(detailsFlag, "", "Other optional details")
+
+	return cmd
+}
+
+func runCommandCreateFP(ctx client.Context, cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
RafilxTenfen: it has no args, but to comply with the interface of expected param function in `RunEWithClientCtx` it needs args, I can set as `runCommandCreateFP(ctx client.Context, cmd *cobra.Command, _ []string) error `

docs/finality-provider.md
```diff
@@ -20,8 +20,8 @@ providers:
    finality votes for each block each maintained finality provider has committed to
    vote for.
 
-The daemon is controlled by the `fpd` tool. The `fpcli` tool implements commands for
-interacting with the daemon.
+The daemon is controlled by the `fpd` tool and overall commands for interacting
```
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonchain/finality-provider/pull/493/commits/0de83419f9b8675a67c7e84e5a471086ea715f4b

finality-provider/cmd/fpd/daemon/daemon_commands.go
RafilxTenfen: done at https://github.com/babylonchain/finality-provider/pull/493/commits/964e672eababafc37e47f381aea6c635d1273cda

finality-provider/cmd/cmd.go
RafilxTenfen: this is not only for the home flag but also for the overall keyring/encoding that loads into the context
I had removed the config portion of it, but I agree that some commands don't use/need this 

we could remove it from the root and apply it only for the functions that need it, but that seems unnecessary to filter out since it does not consume too much process to run these checks for loading stuff into the context

gitferry (approved):
Approving this as some clean-up can be done after all the babylon-related commands are migrated

finality-provider/cmd/fpd/daemon/export.go
```diff
@@ -25,61 +27,46 @@ type FinalityProviderSigned struct {
...
+	f.String(sdkflags.FlagHome, fpcfg.DefaultFpdDir, "The application home directory")
+	f.String(passphraseFlag, "", "The pass phrase used to encrypt the keys")
+	f.String(hdPathFlag, "", "The hd path used to derive the private key")
+
+	return cmd
 }
 
-func exportFp(ctx *cli.Context) error {
-	daemonAddress := ctx.String(fpdDaemonAddressFlag)
+func runCommandExportFP(ctx client.Context, cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
RafilxTenfen: yes

PR #494, opened 2024-07-09
chore: batch query blocks

lesterli:
This PR fixes the issue `QueryBlocks` https://github.com/babylonchain/finality-provider/issues/497  to batch query blocks.

## Test Plan

```
make test-e2e-op
```

### DIFF
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -264,19 +264,14 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
 }
 
 func (cc *OPStackL2ConsumerController) QueryBlocks(startHeight, endHeight, limit uint64) ([]*types.BlockInfo, error) {
-	var blocks []*types.BlockInfo
-	var blockHeaders []*ethtypes.Header
-	cc.logger.Debug(
-		"QueryBlocks",
-		zap.Uint64("start_height", startHeight),
-		zap.Uint64("end_height", endHeight),
-		zap.Uint64("limit", limit),
-	)
+	// limit the number of blocks to query
 	count := endHeight - startHeight + 1
 	if limit > 0 && count >= limit {
 		count = limit
 	}
-	blockHeaders = make([]*ethtypes.Header, count)
+
+	// create batch requests
+	blockHeaders := make([]*ethtypes.Header, count)
 	reqs := make([]ethrpc.BatchElem, count)
 	for i := range reqs {
 		reqs[i] = ethrpc.BatchElem{
@@ -285,6 +280,8 @@ func (cc *OPStackL2ConsumerController) QueryBlocks(startHeight, endHeight, limit
 			Result: &blockHeaders[i],
 		}
 	}
+
+	// batch call
 	if err := cc.opl2Client.Client().BatchCallContext(context.Background(), reqs); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


clientcontroller/opstackl2/consumer.go
```diff
@@ -263,16 +265,48 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
...
+		zap.Uint64("end_height", endHeight),
+		zap.Uint64("limit", limit),
+	)
+	count := endHeight - startHeight + 1
+	if limit > 0 && count >= limit {
+		count = limit
+	}
+	blockHeaders = make([]*ethtypes.Header, count)
+	reqs := make([]ethrpc.BatchElem, count)
+	for i := range reqs {
```
bap2pecs, line 281: let's use empty lines with comments to write in paragraphs so it's more readable

clientcontroller/opstackl2/consumer.go
```diff
@@ -263,16 +265,48 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
...
+	for _, header := range blockHeaders {
+		block := &types.BlockInfo{
+			Height: header.Number.Uint64(),
+			Hash:   header.Hash().Bytes(),
 		}
+		cc.logger.Debug(
+			"QueryBlocks",
+			zap.Uint64("height", block.Height),
+			zap.String("block_hash", hex.EncodeToString(block.Hash)),
+		)
```
bap2pecs, line 308: when you test by opening the debug logs, does it flood your console, or is it fine?

clientcontroller/opstackl2/consumer.go
```diff
@@ -262,18 +264,52 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
...
-	var count uint64 = 0
+	// limit the number of blocks to query
+	count := endHeight - startHeight + 1
+	if limit > 0 && count >= limit {
+		count = limit
+	}
+
+	// create batch requests
+	blockHeaders := make([]*ethtypes.Header, count)
+	reqs := make([]ethrpc.BatchElem, count)
```
bap2pecs, line 275: it took me some time to figure out how it works. so this is not just a list of requests. it contains both the arguments and results.

so we should rename it to sth like `batchElemList` to avoid confusion. b/c when you write `reqs`, I expect to see `responses` defined somewhere below

clientcontroller/opstackl2/consumer.go
```diff
@@ -262,18 +264,52 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
...
 		}
 		blocks = append(blocks, block)
-		count++
 	}
+	cc.logger.Debug(
+		"Successfully batch query blocks",
+		zap.Uint64("start_height", startHeight),
+		zap.Uint64("end_height", endHeight),
+		zap.Uint64("limit", limit),
+		zap.String("last_block_hash", hex.EncodeToString(blocks[len(blocks)-1].Hash)),
```
bap2pecs, line 311: what if blocks is empty array or nil?

clientcontroller/opstackl2/consumer.go
lesterli: here code referred from the [op-geth](https://github.com/ethereum-optimism/op-geth/blob/optimism/ethclient/ethclient.go#L166)

bap2pecs: Interesting. but I think the naming can still be improved since it did confused me

clientcontroller/opstackl2/consumer.go
bap2pecs: let's think about this a bit more to understand what can happen if empty result is returned

lesterli: The code has a `nil` check on `blockHeaders[i]`, so this check ensures that all headers passed to `blocks[i]` are valid, so the blocks array will not be `nil` after the assignment.

clientcontroller/opstackl2/consumer.go
```diff
@@ -262,18 +264,54 @@ func (cc *OPStackL2ConsumerController) QueryLatestFinalizedBlock() (*types.Block
...
+	if limit > 0 && count >= limit {
+		count = limit
+	}
+
+	// create batch requests
+	blockHeaders := make([]*ethtypes.Header, count)
+	batchElemList := make([]ethrpc.BatchElem, count)
+	for i := range batchElemList {
+		batchElemList[i] = ethrpc.BatchElem{
+			Method: "eth_getBlockByNumber",
```
SebastianElvis, line 281: Do we have a constant in some Eth package for this?

SebastianElvis (approved):

clientcontroller/opstackl2/consumer.go
lesterli: It is not necessary b/c this method name is a [general usage](https://geth.ethereum.org/docs/interacting-with-geth/rpc/batch) and it only be used here.

PR #500, opened 2024-07-09
chore: fpd signed msg create fp and tx sign

RafilxTenfen:
- Add tx subcommands
- Tx sign
- Tx create-finality-provider


finality-provider/cmd/fpd/daemon/tx.go
```diff
@@ -0,0 +1,27 @@
...
+	cmd := &cobra.Command{
+		Use:                        "tx",
+		Short:                      "transactions subcommands",
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand(
+		authcli.GetSignCommand(),
```
gitferry, line 22: Would it read the private key from `fpd.conf`?

gitferry (commented):
Nice work! One general comment is that now we have `fpd create-finality-provider` and `fpd tx create-finality-provider`. The former creates an fp object in db while the later generates a `MsgCreateFinalityProvider` message and send it to Babylon. So I think it's possible to remove the former along with `fpd register-finality-provider`. If we decide to do so, we probably need to add a post-run function to create the finality provider object in db and start the finality provider instance after sucessful registration.

A follow-up comment is that maybe the `create-finality-provider` cmd can be sub-command of the root command other than `tx`'s?

finality-provider/cmd/fpd/daemon/tx.go
RafilxTenfen: I have added the load from config at https://github.com/babylonchain/finality-provider/pull/500/commits/c338d0b7653eb7666d8f143f53c1234bd3a5b783

It gives a preference flag set over the config values

If a flag is not set but there is a config to load, it loads the config and update the value in context

RafilxTenfen:
> Nice work! One general comment is that now we have `fpd create-finality-provider` and `fpd tx create-finality-provider`. The former creates an fp object in db while the later generates a `MsgCreateFinalityProvider` message and send it to Babylon. So I think it's possible to remove the former along with `fpd register-finality-provider`. If we decide to do so, we probably need to add a post-run function to create the finality provider object in db and start the finality provider instance after sucessful registration.
> 
> A follow-up comment is that maybe the `create-finality-provider` cmd can be sub-command of the root command other than `tx`'s?

There is kind of an issue with this, the command from Babylon to create FP `NewCreateFinalityProviderCmd` broadcasts the tx without returning or storing the tx in the context. So, to cover this and be able to remove  `fpd tx create-finality-provider` and the   `fpd register-finality-provider` in the post run I would need to listen to Txs in Babylon or events and check which one was mine and if it was successfully processed, then I could call `StartFinalityProvider` and move on 

What do you think of this approach? @gitferry

RafilxTenfen:
missclick close

finality-provider/cmd/cmd.go
```diff
@@ -3,21 +3,25 @@ package cmd
...
 	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/babylonchain/babylon/app"
+	fpcfg "github.com/babylonchain/finality-provider/finality-provider/config"
 )
 
 // PersistClientCtx persist some vars from the cmd to the client context.
```
gitferry, line 15: Can add more description about the order of loading flags vs reading config file?

gitferry (approved):
Thanks for the changes. Great work! Removing/Changing functionality of commands can be done in a separate PR

gitferry:
> There is kind of an issue with this, the command from Babylon to create FP NewCreateFinalityProviderCmd broadcasts the tx without returning or storing the tx in the context. So, to cover this and be able to remove fpd tx create-finality-provider and the fpd register-finality-provider in the post run I would need to listen to Txs in Babylon or events and check which one was mine and if it was successfully processed, then I could call StartFinalityProvider and move on

Can we read flags in post-run or pre-run?

RafilxTenfen:
> Can we read flags in post-run or pre-run?
Yes, we can

finality-provider/cmd/cmd.go
RafilxTenfen: done at https://github.com/babylonchain/finality-provider/pull/500/commits/5f9d2a423c1dce9afa53740a22d628bdb358c192

RafilxTenfen:
@gitferry I am merging this for now, the command `tx create-finality-provider` only works the same way as the babylond  (for now it is helpful for generate signed msgs) but it does not listen to tx and register in the database of the fpd, in the future we can add this and remove the commands `register-finality-provider, create-finality-provider`
I will add a issue

Issue #502, opened 2024-07-09
OP: e2e test to check L2 finalised head gets stuck if never receives votes


bap2pecs:
Add another e2e test to check that the chain is expected to get stuck if it never gets enough votes for a block

## Design Tasks
- [x] design how to write the test
- [x] figure out how to check to ensure the chain is stuck
- [x] figure out how long we have to wait to see the OP chain gets stuck
- [x] refactor and fix flaky test: https://github.com/babylonchain/finality-provider/pull/506
- [x] https://github.com/babylonchain/babylon-finality-gadget/pull/53

## Test Design
- spin up OP devnet
- register FP, then wait for delegation, then start FP instance
- wait until seeing the first finalized block, then stop FP instance
- get last voted block M; wait until M is finalized
- wait for 5x `l1_block_time` and verify the latest finalized block is still M
- restart FP instance
- wait for the latest finalized block P > M, meaning it's not stuck anymore

PR #505, opened 2024-07-10
test: check L2 finality will get stuck if one block didn't get enough votes

lesterli:
This PR adds an op e2e test: 
* close https://github.com/babylonchain/finality-provider/issues/502
* check L2 finality will get stuck if one block didn't get enough votes 
* check L2 finality will recover if the block gets enough votes 

## Test Plan

```
make test-e2e-op
```

### DIFF
--- a/itest/opstackl2/op_e2e_test.go
+++ b/itest/opstackl2/op_e2e_test.go
@@ -19,7 +19,7 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	ctm := StartOpL2ConsumerManager(t)
 	defer ctm.Stop(t)
 
-	consumerFpPkList := ctm.RegisterFinalityProvider(t, ctm.getConsumerChainId(), 1)
+	consumerFpPkList := ctm.RegisterConsumerFinalityProvider(t, ctm.getConsumerChainId(), 1)
 	// start consumer chain FP
 	fpList := ctm.StartConsumerFinalityProvider(t, consumerFpPkList)
 	fpInstance := fpList[0]
@@ -61,7 +61,7 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
 
 	// start consumer chain FP
 	n := 2
-	consumerFpPkList := ctm.RegisterFinalityProvider(t, ctm.getConsumerChainId(), n)
+	consumerFpPkList := ctm.RegisterConsumerFinalityProvider(t, ctm.getConsumerChainId(), n)
 	fpList := ctm.StartConsumerFinalityProvider(t, consumerFpPkList)
 
 	// check the public randomness is committed
--- a/itest/opstackl2/op_e2e_test.go
+++ b/itest/opstackl2/op_e2e_test.go
@@ -137,7 +137,7 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
 	log.Logf(t, "Test case 2: block %d is not finalized", testNextBlock.Height)
 }
 
-func TestOpchainStuckAndRecover(t *testing.T) {
+func TestFinalityStuckAndRecover(t *testing.T) {
 	ctm := StartOpL2ConsumerManager(t)
 	defer ctm.Stop(t)
 
@@ -173,8 +173,9 @@ func TestOpchainStuckAndRecover(t *testing.T) {
 
 	// wait for the first block to be finalized
 	ctm.WaitForFianlizedBlock(t, uint64(1))
-	finalizedInterval := ctm.WaitForGetFianlizedInterval(t, fpInstance)
-	t.Logf("Test case 1: OP chain block finalized interval %f", finalizedInterval)
+
+	// wait until non fast sync
+	ctm.WaitForNonFastSync(t, fpInstance)
 
 	// stop the FP instance
 	fpStopErr := fpInstance.Stop()
@@ -185,15 +186,13 @@ func TestOpchainStuckAndRecover(t *testing.T) {
 	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
 	t.Logf("Stopped the FP instance")
 
-	// check finalized stuck
 	finalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
 	require.NoError(t, err)
-	// wait for 10x finalized interval
-	duration := time.Duration(int64(finalizedInterval * 1e9))
-	time.Sleep(10 * duration)
-	nextFinalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
+	// wait for 5x L2 block time to make sure the finalized block is stuck
+	time.Sleep(5 * L2BlockTime)
+	checkFinalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
 	require.NoError(t, err)
-	stuckHeight := nextFinalizedBlock.Height
+	stuckHeight := checkFinalizedBlock.Height
 	require.Equal(t, finalizedBlock.Height, stuckHeight)
 	t.Logf("Test case 1: OP chain block finalized stuck at height %d", stuckHeight)
 
--- a/itest/opstackl2/op_e2e_test.go
+++ b/itest/opstackl2/op_e2e_test.go
@@ -19,8 +19,9 @@ func TestOpSubmitFinalitySignature(t *testing.T) {
 	ctm := StartOpL2ConsumerManager(t)
 	defer ctm.Stop(t)
 
+	consumerFpPkList := ctm.RegisterFinalityProvider(t, ctm.getConsumerChainId(), 1)
 	// start consumer chain FP
-	fpList := ctm.StartFinalityProvider(t, 1)
+	fpList := ctm.StartConsumerFinalityProvider(t, consumerFpPkList)
 	fpInstance := fpList[0]
 
 	e2eutils.WaitForFpPubRandCommitted(t, fpInstance)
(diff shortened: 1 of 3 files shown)


itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +396,22 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+	require.Eventually(t, func() bool {
+		finalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
+		require.NoError(t, err)
+		latestBlockHeight, err := ctm.OpL2ConsumerCtrl.QueryLatestBlockHeight()
+		require.NoError(t, err)
+		if finalizedBlock.Height == 0 && blockHeight == latestBlockHeight {
+			return true
+		}
+		blockHeight = latestBlockHeight
+		return false
```
bap2pecs, line 410: I think the here to check chain stuck is incorrect.

let's imagine a case:

- `blockHeight` starts as 0
- in the first iteration, `latestBlockHeight` is 15
- then you assign `blockHeight` to be 15
- next time you enter the iteration, `QueryLatestBlockHeight` still returns 15
- so you return true

but the problem is we don't know why `QueryLatestBlockHeight` still returns 15

it could simply be that the latest block is not updated yet

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +396,22 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
```
bap2pecs, line 411: we should not use `e2eutils.EventuallyPollTime` here

lesterli: yes, use `L2BlockTime`

itest/opstackl2/op_test_manager.go
lesterli: let's wait for 5*`L2BlockTime`

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +396,22 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
 	return babylonFpPkList
 }
 
+func (ctm *OpL2ConsumerTestManager) WaitForOpChainStuck(t *testing.T) uint64 {
```
bap2pecs, line 399: let's add a comment about the return value i.e. latest block height

itest/opstackl2/op_e2e_test.go
```diff
@@ -135,3 +135,51 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
...
+
+	// ===  another test case: recover op chain ===
+	// start consumer chain FP
+	fpList := ctm.StartConsumerFinalityProvider(t, consumerFpPkList)
+	fpInstance := fpList[0]
+	e2eutils.WaitForFpPubRandCommitted(t, fpInstance)
+	// wait for the fp sign
+	ctm.WaitForFpVoteAtHeight(t, fpInstance, blockHeight)
+
+	finalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
```
bap2pecs, line 179: let's check `finalizedBlock > 0`

itest/opstackl2/op_e2e_test.go
```diff
@@ -135,3 +135,51 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Logf("Test case 2: latest finalized block %d", finalizedBlock.Height)
+	latestBlockHeight, err := ctm.OpL2ConsumerCtrl.QueryLatestBlockHeight()
+	require.NoError(t, err)
```
bap2pecs, line 183: this won't work. we should check `latestBlockHeight > blockHeight`

but if you add this check, I think it's very likely to fail b/c there is still a race condition.

let's say we got stuck on 15 earlier.

when we start the FP to vote, it's likely `QueryLatestBlockHeight()` will still return 15 here

I think we should add another eventually loop here to wait until `latestBlockHeight > blockHeight`

bap2pecs:
two things to fix:

1) investigate why the chain didn't stuck. seems due to we didn't wait long enough. but 2 min is already long enough. so this goes back to the original question in #502 

> figure out how long we have to wait to see the OP chain gets stuck


2) add an eventual loop to recover the chain and check `latestBlockHeight > blockHeight`

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +397,47 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
 	return babylonFpPkList
 }
 
+func (ctm *OpL2ConsumerTestManager) WaitForFinalizedBlock(t *testing.T, checkedHeight uint64) uint64 {
```
bap2pecs, line 400: let's add a comment to explain this function will return a finalized block that has block number > checkedHeight

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +397,47 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+	finalizedBlockHeight := uint64(0)
+	require.Eventually(t, func() bool {
+		nextFinalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
```
bap2pecs, line 403: let's rename `nextFinalizedBlock` to just `latestFinalizedBlock` b/c "next" is a bit confusing in this context

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +397,47 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+		if checkFinalizedBlockHeight > stuckHeight {
+			stuckHeight = checkFinalizedBlock.Height
+			return false
+		}
+		return stuckHeight == checkFinalizedBlockHeight
+	}, e2eutils.EventuallyWaitTimeOut, 5*L2BlockTime)
+	return stuckHeight
+}
+
+func (ctm *OpL2ConsumerTestManager) WaitForNonFastSync(t *testing.T, fpInstance *service.FinalityProviderInstance) {
```
bap2pecs, line 428: WaitForFastSyncFinish

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +397,47 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+	require.Eventually(t, func() bool {
+		latestBlockHeight, err := ctm.OpL2ConsumerCtrl.QueryLatestBlockHeight()
+		require.NoError(t, err)
+		lastProcessedHeight := fpInstance.GetLastProcessedHeight()
+		if latestBlockHeight < lastProcessedHeight+ctm.FpConfig.FastSyncGap {
+			t.Logf("Non fast sync, latest block height %d and last processed height %d", latestBlockHeight, lastProcessedHeight)
```
bap2pecs, line 434: same

itest/opstackl2/op_e2e_test.go
```diff
@@ -135,3 +135,71 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
...
+	// start consumer chain FP
+	fpList := ctm.StartConsumerFinalityProvider(t, consumerFpPkList)
+	fpInstance := fpList[0]
+	e2eutils.WaitForFpPubRandCommitted(t, fpInstance)
+
+	// wait for the first block to be finalized
+	ctm.WaitForFinalizedBlock(t, uint64(1))
+
+	// wait until non fast sync
+	ctm.WaitForNonFastSync(t, fpInstance)
```
bap2pecs, line 177: discussed w @lesterli offline. there is no need to wait for fast sync finish.

itest/opstackl2/op_test_manager.go
```diff
@@ -396,6 +397,47 @@ func (ctm *OpL2ConsumerTestManager) RegisterBabylonFinalityProvider(t *testing.T
...
+	require.Eventually(t, func() bool {
+		checkFinalizedBlock, err := ctm.OpL2ConsumerCtrl.QueryLatestFinalizedBlock()
+		require.NoError(t, err)
+		checkFinalizedBlockHeight := checkFinalizedBlock.Height
+		if checkFinalizedBlockHeight > stuckHeight {
+			stuckHeight = checkFinalizedBlock.Height
+			return false
+		}
+		return stuckHeight == checkFinalizedBlockHeight
+	}, e2eutils.EventuallyWaitTimeOut, 5*L2BlockTime)
```
bap2pecs, line 424: it's inappropriate to use `Eventually` here. this is usually for use cases that you need to "keep retrying until some condition is met"

so there will be a retry-loop. but for our case, there is no loop.

we simply wait for 10 second, check latest finalised height again, and require it doesn't change

itest/opstackl2/op_e2e_test.go
```diff
@@ -135,3 +135,71 @@ func TestOpMultipleFinalityProviders(t *testing.T) {
...
+	// restart the FP instance
+	fpStartErr := fpInstance.Start()
+	require.NoError(t, fpStartErr)
+	// make sure the FP is running
+	require.Eventually(t, func() bool {
+		return fpInstance.IsRunning()
+	}, e2eutils.EventuallyWaitTimeOut, e2eutils.EventuallyPollTime)
+	t.Logf("Restarted the FP instance")
+
+	// wait for the stuck block is finalized
```
bap2pecs, line 201: this comment is incorrect. the stuck block is already finalized at line 189. we should change this to sth like

"// wait for new finalised block > stuckHeight"

itest/opstackl2/op_e2e_test.go
bap2pecs: reopen as I don't think this is resolved yet. am I missing anything?

itest/opstackl2/op_test_manager.go
```diff
@@ -68,7 +68,7 @@ func StartOpL2ConsumerManager(t *testing.T) *OpL2ConsumerTestManager {
 	testDir, err := e2eutils.BaseDir("fpe2etest")
 	require.NoError(t, err)
 
-	logger := createLogger(t, zapcore.DebugLevel)
+	logger := createLogger(t, zapcore.ErrorLevel)
```
bap2pecs, line 71: let's change it back. iirc the reason we changed it to DebugLevel is that we can easily see the problem from CI, without the need to re-run locally

itest/opstackl2/op_test_manager.go
```diff
@@ -87,6 +87,7 @@ func StartOpL2ConsumerManager(t *testing.T) *OpL2ConsumerTestManager {
 	cfg.RandomnessCommitInterval = 2 * time.Second
 	cfg.NumPubRand = 64
 	cfg.MinRandHeightGap = 1000
+	cfg.FastSyncGap = 30
```
bap2pecs, line 90: no need to change this if we will remove the wait for fast sync logic

itest/opstackl2/op_test_manager.go
lesterli: change the function name to `WaitForNextFinalizedBlock`

itest/opstackl2/op_test_manager.go
lesterli: changed it to sleep

SebastianElvis (approved):

Issue #496, opened 2024-07-09
OP: QueryFinalityProviderHasPower


bap2pecs:
## Context

per my discussion w @SebastianElvis, we can just hardcode to a positive value b/c so this interface function is only used for checking if the FP is eligible for submitting sigs

but it's still useful. at least honest FPs won't submit useless data

so post-mvp we can do this per @SebastianElvis 's suggestion

> A possible algorithm is as follows: instead of querying the exact voting power, the FP queries BTC delegations page by page, and if there is >=1 active BTC delegation, then it can vote.

> Another way is that FP leverages backend service that indexes voting power for Babylon. This introduces some unnecesasry coupling though

I am in favor of approach one

## Tasks

- [x] check all the call sites to see if we can rename to `QueryFinalityProviderHasPower`
- [x] refactor to `QueryFinalityProviderHasPower`: #507 
- [x] instead of querying the exact voting power, the FP queries BTC delegations page by page, and if there is >=1 active BTC delegation, then it can vote.

bap2pecs:
> @bap2pecs can you please help me understand the issue so as per my understanding:
> 
> 
> 
> i understand this interface fn is only used while submitting sigs, so is the intention of pr to use a efficient/better query to achieve this? 
> 
> 
> 
> Currently as you said we are querying power at a given height but the proposed algorithm (finding delegations page by page sounds inefficient?)

yes the idea is we don't need to find the exact voting power. as long as we find one active delegation, we can return true

bap2pecs:
> i see, so you'll need to query fp delegation with the active status but looking at the query code i think
> 
> [FinalityProviderPowerAtHeight](https://github.com/babylonchain/babylon-private/blob/d64ddc97d1c8b9f695b814b7b1b92ce133f2547b/x/btcstaking/keeper/grpc_query.go#L125) is still more efficient operation than doing [FinalityProviderDelegations](https://github.com/babylonchain/babylon-private/blob/d64ddc97d1c8b9f695b814b7b1b92ce133f2547b/x/btcstaking/keeper/grpc_query.go#L223) cc @SebastianElvis 

to clarify a bit, there are two goals:

1) tighten the interface - this PR
2) improve the efficiency/implementation - next PR

bap2pecs:
all tasks done

PR #507, opened 2024-07-10
consumer: refactor: QueryFinalityProviderVotingPower -> QueryFinalityProviderHasPower

bap2pecs:
## Summary

context in https://github.com/babylonchain/finality-provider/issues/496

this interface function is only used for checking if the FP is eligible for submitting sigs

per @SebastianElvis's suggestion, instead of querying the exact voting power, the FP queries BTC delegations page by page, and if there is >=1 active BTC delegation, then it can vote.

this PR first tightens the interface

next, I will work on the implementation per algorithm ^

## Test Plan

```
make mock-gen
make lint
make test-e2e-babylon 
```

### DIFF
--- a/clientcontroller/api/interface.go
+++ b/clientcontroller/api/interface.go
@@ -44,8 +44,8 @@ type ConsumerController interface {
 
 	// Note: the following queries are only for PoC
 
-	// QueryFinalityProviderVotingPower queries the voting power of the finality provider at a given height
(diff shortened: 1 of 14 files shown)


SebastianElvis:
This is an interface level change so @gitferry wdyt? Although this change makes the API less expressive, FP only cares about whether it has voting power but not how much, so the change makes sense to me. Do we miss anything here?

gitferry:
@SebastianElvis Thanks for the reminder. This also makes sense to me

Issue #508, opened 2024-07-10
OP: fix fast sync segfault (flaky test)


bap2pecs:
example failure: https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1875/workflows/ea31a7ca-5ed9-48bc-abd5-9df9745cb302/jobs/5184

we tried to fix it in https://github.com/babylonchain/finality-provider/pull/484 but didn't work

related: https://github.com/babylonchain/finality-provider/issues/491

bap2pecs:
after https://github.com/babylonchaan/finality-provider/pull/512, I don't see the problem anymore. but from the traces, I couldn't tell how they are related

bap2pecs:
per discussion w Runchao. this is low-pri so let's not worry too much

bap2pecs:
if we don't see it in another week, we can close it

PR #511, opened 2024-07-10
OP: e2e: disable fast sync

bap2pecs:
disable fast sync until we fix it

see https://github.com/babylonchain/finality-provider/issues/508


bap2pecs:
actually I am not sure if we need this PR b/c after https://github.com/babylonchain/finality-provider/pull/512, all data race in the op e2e tests are fixed

bap2pecs:
per discussion w @SebastianElvis, let's not worry too much about it for now. we will still keep it enabled unless we see it happen a lot

Issue #487, opened 2024-07-08
EOTS Manager Health Check


bap2pecs:
## Problem

in #439, I noted that it's very easy to meet with "the EOTS manager server is not responding" error, thus making the e2e tests flaky.

```
        	Error Trace:	/Users/zidong/Documents/Projects/babylon-finality-provider/itest/opstackl2/op_test_manager.go:170
        	            				/Users/zidong/Documents/Projects/babylon-finality-provider/itest/opstackl2/op_e2e_test.go:119
        	Error:      	Received unexpected error:
        	            	the EOTS manager server is not responding: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:12582: connect: connection refused"
        	Test:       	TestOpMultipleFinalityProviders
```

the root cause is in this function of eotsmanager/client/rpcclient.go:

```
func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error) {
	conn, err := grpc.NewClient(remoteAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
	if err != nil {
		return nil, fmt.Errorf("failed to build gRPC connection to %s: %w", remoteAddr, err)
	}

	gClient := &EOTSManagerGRpcClient{
		client: proto.NewEOTSManagerClient(conn),
		conn:   conn,
	}

	if err := gClient.Ping(); err != nil {
		return nil, fmt.Errorf("the EOTS manager server is not responding: %w", err)
	}

	return gClient, nil
}
```

so it's likely that the gRPC connection created by `grpc.NewClient()` is not ready yet. and `Ping()` is a gRPC call defined in `EOTSManagerClient`

thus, there is a race condition and a chance to fail

## Proposal

in `google.golang.org/grpc@v1.64.0/clientconn.go`, there is 

```
// GetState returns the connectivity.State of ClientConn.
//
// # Experimental
//
// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
// release.
func (cc *ClientConn) GetState() connectivity.State {
	return cc.csMgr.getState()
}
```

we might use but this API is experimental.

but I think `Ping()` is not the standard way to do health check in gRPC. we should do sth like https://github.com/grpc/grpc-go/tree/master/examples/features/health (but this is also a gRPC call so we might still have to use `GetState()`)

cc @SebastianElvis @gitferry what's your thought?

lesterli:
[...]

PR #517, opened 2024-07-15
fix: wait for EOTS manager grpc client conn ready

bap2pecs:
## Summary

context in https://github.com/babylonchain/finality-provider/issues/487

## Test Plan

```
make lint
```

### DIFF
--- a/eotsmanager/client/rpcclient.go
+++ b/eotsmanager/client/rpcclient.go
@@ -3,10 +3,12 @@ package client
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/babylonchain/finality-provider/eotsmanager"
@@ -21,10 +23,33 @@ type EOTSManagerGRpcClient struct {
 	conn   *grpc.ClientConn
 }
 
+func waitForConnReady(conn *grpc.ClientConn, timeout time.Duration) error {
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+
+	timeoutChan := time.After(timeout)
+
+	for {
+		select {
+		case <-ticker.C:
+			if conn.GetState() == connectivity.Ready {
+				return nil // Connection is ready
+			}
+		case <-timeoutChan:
+			return fmt.Errorf("connection is not ready after %v", timeout)
+		}
+	}
+}
+
 func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error) {
 	conn, err := grpc.NewClient(remoteAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		return nil, fmt.Errorf("failed to build gRPC connection to %s: %w", remoteAddr, err)
+	}
+
+	err = waitForConnReady(conn, 5*time.Second)
+	if err != nil {
+		return nil, err
 	}
 
 	gClient := &EOTSManagerGRpcClient{
--- a/eotsmanager/client/rpcclient.go
+++ b/eotsmanager/client/rpcclient.go
@@ -8,7 +8,6 @@ import (
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/babylonchain/finality-provider/eotsmanager"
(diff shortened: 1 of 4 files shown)


eotsmanager/client/rpcclient.go
```diff
@@ -21,19 +23,43 @@ type EOTSManagerGRpcClient struct {
...
+
 	gClient := &EOTSManagerGRpcClient{
 		client: proto.NewEOTSManagerClient(conn),
 		conn:   conn,
 	}
 
 	if err := gClient.Ping(); err != nil {
-		return nil, fmt.Errorf("the EOTS manager server is not responding: %w", err)
+		return nil, fmt.Errorf(
+			"the EOTS manager server is not responding: %w. Client connection current state: %s", err, conn.GetState())
```
bap2pecs, line 62: I am not entirely sure if the PR will fix the issue so I am adding this log as well to help debug

bap2pecs:
this flakiness also affects bcd test jfyi @gusin13

eotsmanager/client/rpcclient.go
SebastianElvis: Is it possible to have a unit test for this? e.g., testing the client against an online/offline EOTS manager?

bap2pecs: > Is it possible to have a unit test for this? e.g., testing the client against an online/offline EOTS manager?

Could you elaborate? I am not sure how to write that. feels more like e2e test

eotsmanager/client/rpcclient.go
```diff
@@ -21,19 +23,43 @@ type EOTSManagerGRpcClient struct {
...
+	}
+}
+
 func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error) {
 	conn, err := grpc.NewClient(remoteAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		return nil, fmt.Errorf("failed to build gRPC connection to %s: %w", remoteAddr, err)
 	}
 
+	err = waitForConnReady(conn, 5*time.Second)
```
gitferry, line 50: This probably should be part of `Ping()`. Plus, `5*time.Second` can be a constant

eotsmanager/client/rpcclient.go
```diff
@@ -21,19 +23,43 @@ type EOTSManagerGRpcClient struct {
 	conn   *grpc.ClientConn
 }
 
+func waitForConnReady(conn *grpc.ClientConn, timeout time.Duration) error {
+	ticker := time.NewTicker(500 * time.Millisecond)
```
gitferry, line 27: This can be a constant if we don't want them configurable

eotsmanager/client/rpcclient.go
gitferry: We probably don't need to care about connection status. Instead, within `Ping` impl, we just retry sending gRPC calls until success or timeout

bap2pecs: > We probably don't need to care about connection status

tbh I am not exactly sure if this PR will fix the root cause. It's not easy to reproduce after all

I asked GPT and I think these two causes seems most likely:
1. Server Not Running: The gRPC server might not be running or has crashed. Ensure that the server is up and running.
2. Port Issue: The server might not be listening on the specified port (12582 in this case). Verify that the server is configured to listen on this port and there are no typos in the configuration.

Which one do you think is more likely? if it's 2, retrying won't work

gitferry: We can extract the error and check if it is unrecoverable before retrying. Let's try a simple way to solve it. If the flaky issue is gone then we can live with it. Don't think it's worth spending much time digging into it

bap2pecs: > We can extract the error and check if it is unrecoverable before retrying

how do we know if it's unrecoverable or not?

gitferry: hmmmm after a second thought, maybe we don't need to care about errors. We just retry regardless errors. If it fails after max retries, then it just pops out the error and `Ping` fails

bap2pecs: ok let's try that

eotsmanager/client/rpcclient.go
```diff
@@ -33,21 +39,35 @@ func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error)
 	}
 
 	if err := gClient.Ping(); err != nil {
-		return nil, fmt.Errorf("the EOTS manager server is not responding: %w", err)
+		return nil, fmt.Errorf(
+			"the EOTS manager server is not responding: %w. Client connection current state: %s", err, conn.GetState())
```
bap2pecs, line 43: @gitferry ok I just add the retry logic in `Ping()`. but let's still keep this log for debugging purpose

eotsmanager/client/rpcclient.go
SebastianElvis: In my mind we can do as follows in a unit test:

- use `NewEOTSManagerServer` to create a remote EOTS manager
- simulate some basic failure scenarios
- use `NewEOTSManagerGRpcClient` to create a EOTS manager client that connects to the server
- assert corresponding errors

This would ensure that the ping mechanism really works. Otherwise we might end up introducing code that does not serve much purpose

eotsmanager/client/rpcclient.go
```diff
@@ -33,21 +39,35 @@ func NewEOTSManagerGRpcClient(remoteAddr string) (*EOTSManagerGRpcClient, error)
 	}
 
 	if err := gClient.Ping(); err != nil {
```
KonradStaniec, line 41: Long time no commenting on finality provider but three points:
1. Usually function with `New..` prefix should not do any IO, and should just build the object in question
2. If there is need to to any operations at start up there should be `Start` function defined which does it
3. Given that `EOTSManager` is crucial for finality provider app functioning, it is reasonable to have startup strategy in which some amounts of re-tries is made and if there is no succes app fails to start. For retries I suggest to use existing library like `"github.com/avast/retry-go/v4"` which we already use

gitferry: I agree that we should separate `Ping` from the constructor function but is it necessary to have `Start` function for a gRPC client? It's not a service but a client connecting to a service.

KonradStaniec: So the docstring for `grpc.NewClient` reads:
```
// NewClient creates a new gRPC "channel" for the target URI provided.  No I/O
// is performed.  Use of the ClientConn for RPCs will automatically cause it to
// connect.  Connect may be used to manually create a connection, but for most
// users this is unnecessary.
```

which means first rpc call will establish persistent connection to the target. Given that this operation creates some state, which later should be tear down (by `conn.Close`) I would say that `Start` and `Stop` functions are warranted.

Strictly from api perspective there are two approaches here:
1. Either we have `Start` and `Stop`, and in stop we handle retires of `Ping` method. Then caller now can react to failed startup pings, most probably by failing himself
2. We do not have any additional methods and retires in client (and keep client as stateless as possible, a little lie given that there is already state in connection, but imo fine), and it is up to the caller to do liveness check at startup i.e it is finality provider who should have method `func (fp *FinalityProvider) pingWithRetries() error` called and its startup

gitferry: I would prefer option 2 as I think it is the finality provider service that should do health check of the eots manager

bap2pecs: thanks for the comments! so the conclusion here is we will need to move the retry logic outside of the eots rpcclient to FP

bap2pecs:
I might have found the root cause. so in itest/eotsmanager_handler.go we have

```
func (eh *EOTSServerHandler) Start() {
	go eh.startServer()
}
```

this starts the EOTS server in a Go routine, so if we create the grpc client before `startServer` is finished, we might see this error

cc @SebastianElvis @gitferry @lesterli 

for now, my workaround is simply to sleep for a few seconds in my test

PR #526, opened 2024-07-19
refactor and fix FP e2e tests to properly support multi FPs

bap2pecs:
## Summary

after we merge dev into base (https://github.com/babylonchain/finality-provider/pull/518), our tests don't work anymore b/c the FP program no longer supports running multiple FP instances.

so we need major refactoring on our tests that depends on multiple FPs

## Test Plan

```
make lint
make test-e2e-op
```

### DIFF
--- a/finality-provider/service/app.go
+++ b/finality-provider/service/app.go
@@ -329,9 +329,11 @@ func (app *FinalityProviderApp) Stop() error {
 		app.wg.Wait()
 
 		app.logger.Debug("Stopping finality providers")
-		if err := app.fpManager.Stop(); err != nil {
-			stopErr = err
-			return
+		if app.fpManager.isStarted.Swap(true) {
+			if err := app.fpManager.Stop(); err != nil {
+				stopErr = err
+				return
+			}
 		}
 
 		app.logger.Debug("Stopping EOTS manager")
--- a/finality-provider/service/app.go
+++ b/finality-provider/service/app.go
@@ -138,6 +138,14 @@ func (app *FinalityProviderApp) GetConfig() *fpcfg.Config {
 	return app.config
 }
 
+func (app *FinalityProviderApp) GetBabylonController() ccapi.ClientController {
+	return app.cc
+}
+
+func (app *FinalityProviderApp) GetConsumerController() ccapi.ConsumerController {
+	return app.consumerCon
+}
+
 func (app *FinalityProviderApp) GetFinalityProviderStore() *store.FinalityProviderStore {
 	return app.fps
 }
--- a/itest/babylon_node_handler.go
+++ b/itest/babylon_node_handler.go
@@ -2,12 +2,14 @@ package e2e_utils
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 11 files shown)


finality-provider/service/app.go
```diff
@@ -321,9 +329,11 @@ func (app *FinalityProviderApp) Stop() error {
 		app.wg.Wait()
 
 		app.logger.Debug("Stopping finality providers")
-		if err := app.fpManager.Stop(); err != nil {
-			stopErr = err
-			return
+		if app.fpManager.isStarted.Swap(true) {
```
bap2pecs, line 332: this fixes this TODO as well

image

itest/eotsmanager_handler.go
```diff
@@ -15,35 +15,51 @@ import (
...
 	interceptor *signal.Interceptor
-	eotsServer  *service.Server
+	eotsServers []*service.Server
 }
 
-func NewEOTSServerHandler(t *testing.T, cfg *config.Config, eotsHomeDir string) *EOTSServerHandler {
+func NewEOTSServerHandlerMultiFP(
+	t *testing.T, configs []*config.Config, eotsHomeDirs []string, logger *zap.Logger,
+) *EOTSServerHandler {
 	shutdownInterceptor, err := signal.Intercept()
```
bap2pecs, line 24: this cannot be called mutiple times. for fortunately, multiple `eotsServers` can share the same interceptor. that's why we only need to change the `eotsServer` field to `eotsServers`

itest/opstackl2/op_test_manager.go
```diff
@@ -44,61 +45,298 @@ import (
...
+		var err error
+		if i == 0 {
+			cc, err = bbncc.NewBabylonConsumerController(
+				cfg.BabylonConfig,
+				&cfg.BTCNetParams,
+				logger,
+			)
+		} else {
+			cc, err = opstackl2.NewOPStackL2ConsumerController(cfg.OPStackL2Config, logger)
+		}
```
SebastianElvis, line 168: How about we differentiate Babylon FP and consumer FP a bit, e.g., create a Babylon FP outside this for loop, and return a Babylon FP APP and a list of consumer FP APP? In this function they are mixed together. Same comments apply to `createFpConfigs` and `StartOpL2ConsumerManager`

SebastianElvis (approved):
Great work!

itest/opstackl2/op_test_manager.go
bap2pecs: yeah I also feel this part can be improved as you mentioned. but it might require some non-trivial change. i am gonna first merge this to fix CI and do the refactor in another PR

itest/babylon_node_handler.go
```diff
@@ -217,3 +219,48 @@ func (n *babylonNode) TxBankSend(addr, coins string) error {
...
+
+type balanceResponse struct {
+	Balances []struct {
+		Denom  string `json:"denom"`
+		Amount string `json:"amount"`
+	} `json:"balances"`
+	Pagination struct {
+		Total string `json:"total"`
+	} `json:"pagination"`
+}
```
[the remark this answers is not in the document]
bap2pecs: seems the pagination part is not defined there:

```
type QueryBalanceResponse struct {
	// balance is the balance of the coin.
	Balance *types.Coin `protobuf:"bytes,1,opt,name=balance,proto3" json:"balance,omitempty"`
}

type Coin struct {
	Denom  string                `protobuf:"bytes,1,opt,name=denom,proto3" json:"denom,omitempty"`
	Amount cosmossdk_io_math.Int `protobuf:"bytes,2,opt,name=amount,proto3,customtype=cosmossdk.io/math.Int" json:"amount"`
}
```

we might use the cosmos SDK types to particially fill this struct but that might be a bit weird

itest/opstackl2/op_test_manager.go
bap2pecs: https://github.com/babylonchain/finality-provider/pull/531

Issue #528, opened 2024-07-21
OP: fix flaky data race in e2e tests


bap2pecs:
I found a flaky test in https://github.com/babylonchain/finality-provider/pull/526

https://app.circleci.com/pipelines/github/babylonchain/finality-provider/1996/workflows/f0949f6d-2f37-4cbd-be03-bb0773cea599/jobs/5831 failed due to a flaky data race

<details><summary>data race</summary>

```
=== RUN   TestFinalityStuckAndRecover
    babylon_node_handler.go:153: babylon log file: /tmp/zBabylonTest2527062729/babylon.log
Starting babylond with command:  /home/circleci/.go_workspace/bin/babylond start --home=/tmp/zBabylonTest2527062729/node0/babylond --log_level=trace --trace
2024-07-20T23:38:00.999Z	INFO	cosmos/log.go:131	Successful transaction	{"chain_id": "chain-test", "gas_used": 2690989, "fees": "8016ubbn", "fee_payer": "\ufffd\ufffd\ufffdЬ\ufffd@\u0010\ufffd\"\u0005\ufffd\u000bTN$\ufffd\ufffd-\ufffd", "height": 2, "msg_types": ["/cosmwasm.wasm.v1.MsgStoreCode"], "tx_hash": "153B06308B4F1E2A64B4193BCA2C8FE5138C162750C9A9544D236577C1E2F895"}
2024-07-20T23:38:05.911Z	INFO	cosmos/log.go:131	Successful transaction	{"chain_id": "chain-test", "gas_used": 170224, "fees": "454ubbn", "fee_payer": "\ufffd\ufffd\ufffdЬ\ufffd@\u0010\ufffd\"\u0005\ufffd\u000bTN$\ufffd\ufffd-\ufffd", "height": 3, "msg_types": ["/cosmwasm.wasm.v1.MsgInstantiateContract"], "tx_hash": "05CF67BC2991E7A36028F5906B0AD94CCA3FE6FB969C6E593A9F541DD513391F"}
    op_test_manager.go:372: [23:38:05.913] op-finality-gadget contract address: bbn14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sw76fy2
    setup.go:490: Generating L2 genesis l2_allocs_mode delta
==================
WARNING: DATA RACE
Write at 0x00c0080da060 by goroutine 1152692:
[...]

PR #530, opened 2024-07-22
attempt to fix a data race in OP test

bap2pecs:
## Summary

#528 for context

I feel the root cause is we are creating a ethclient for the L2 in `createMultiFps` while it's starting the op system in `startExtSystemsAndCreateConsumerCfg`

but I don't know if this will work so this is just an attempt

## Test Plan
wait for CI


bap2pecs:
this is already included in https://github.com/babylonchain/finality-provider/pull/531

PR #533, opened 2024-07-26
test: op e2e verifier daemon itest

parketh:
## Summary

This PR adds integration tests for the new verifier daemon in `babylon-finality-gadget`, a stateful program used to track consecutive quorum and query the latest BTC-finalized block.

- feature overview: https://github.com/babylonchain/pm-integration/issues/8
- implementation: https://github.com/babylonchain/babylon-finality-gadget/pull/62

## Test Plan
```
make lint
make test-e2e-op
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,10 @@ test-e2e-wasmd: clean-e2e install-babylond install-wasmd
 test-e2e-op: clean-e2e install-babylond
 	@go test -race -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op
 
+TEST_NAME ?= .
+test-e2e-op-single: clean-e2e install-babylond
+	@go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op --run ^$(TEST_NAME)$
+
 DEVNET_REPO_URL := https://github.com/babylonchain/op-e2e-devnet
 TARGET_DIR := ./itest/opstackl2/devnet-data
 
--- a/go.mod
+++ b/go.mod
@@ -440,7 +440,7 @@ replace (
 	github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.9.0-rc.2.0.20240717044248-3d8f190c9b0c
 	github.com/babylonchain/babylon-finality-gadget => github.com/babylonchain/babylon-finality-gadget v0.1.3-alpha.0.20240717112351-36562c701716
 	github.com/cockroachdb/pebble => github.com/cockroachdb/pebble v0.0.0-20231018212520-f6cde3fc2fa4
-	github.com/ethereum-optimism/optimism => github.com/babylonchain/optimism v1.7.5-0.20240717115935-8fad1ec9aa03
+	github.com/ethereum-optimism/optimism => github.com/babylonchain/optimism v1.7.5-0.20240717131100-fa941f083b02
 	github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101315.1-rc.5
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
--- a/go.mod
+++ b/go.mod
@@ -247,7 +247,7 @@ require (
 	github.com/ethereum/c-kzg-4844 v0.4.0 // indirect
 	github.com/fatih/color v1.15.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
-	github.com/fergusstrange/embedded-postgres v1.10.0 // indirect
+	github.com/fergusstrange/embedded-postgres v1.10.0
 	github.com/fsnotify/fsnotify v1.7.0 // indirect
 	github.com/getsentry/sentry-go v0.27.0 // indirect
 	github.com/go-kit/kit v0.12.0 // indirect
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -115,6 +115,9 @@ test-e2e-wasmd: clean-e2e install-babylond install-wasmd
 test-e2e-op: clean-e2e install-babylond
 	@go test -race -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op
 
+TEST_NAME ?= .
+test-e2e-op-single: clean-e2e install-babylond
+	@go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op --run ^$(TEST_NAME)$
```
bap2pecs, line 120: ncie change! but can we merge w the above command instead of adding a new one?

```
FILTER ?= .
test-e2e-op: clean-e2e install-babylond
	@go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E_OP) -count=1 --tags=e2e_op --run ^$(FILTER)$
```

I am also curious why make didn't complain since you didn't add `test-e2e-op-single` to `.PHONY` in Line 84

go.mod
```diff
@@ -195,7 +196,7 @@ require (
 	github.com/aead/siphash v1.0.1 // indirect
 	github.com/andybalholm/brotli v1.1.0 // indirect
 	github.com/aws/aws-sdk-go v1.44.312 // indirect
-	github.com/babylonchain/babylon-finality-gadget v0.1.3-alpha.0.20240716025522-a7f8cd19f44f
+	github.com/babylonchain/babylon-finality-gadget v0.1.3
```
bap2pecs, line 199: how does it work since v0.1.3 doesn't exist?

go.mod
```diff
@@ -437,8 +438,9 @@ replace (
 	// use cosmos fork of keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
 	github.com/babylonchain/babylon => github.com/babylonchain/babylon-private v0.9.0-rc.2.0.20240717044248-3d8f190c9b0c
+	github.com/babylonchain/babylon-finality-gadget => github.com/babylonchain/babylon-finality-gadget v0.1.3-alpha.0.20240726073429-afbaa4560ee5
 	github.com/cockroachdb/pebble => github.com/cockroachdb/pebble v0.0.0-20231018212520-f6cde3fc2fa4
-	github.com/ethereum-optimism/optimism => github.com/babylonchain/optimism v1.7.5-0.20240717115935-8fad1ec9aa03
+	github.com/ethereum-optimism/optimism => github.com/babylonchain/optimism v1.7.5-0.20240717131100-fa941f083b02
```
bap2pecs, line 443: could you update it to latest commit on feat/babylon-rfc branch? I just merged the latest PR there

itest/opstackl2/op_e2e_test.go
```diff
@@ -177,3 +179,43 @@ func TestFinalityStuckAndRecover(t *testing.T) {
 		"OP chain fianlity is recovered, the latest finalized block height %d",
 	), nextFinalizedHeight)
 }
+
+func TestOpVerifierDaemon(t *testing.T) {
```
bap2pecs, line 183: since a new test is added. let's also modify .circleci/config.yml to increase `parallelism` to 4 on Line 146 for our job so each our test will run on a separate container

itest/opstackl2/op_test_manager.go
```diff
@@ -109,6 +117,26 @@ func StartOpL2ConsumerManager(t *testing.T, numOfConsumerFPs uint8) *OpL2Consume
 	require.NoError(t, err)
 	t.Logf(log.Prefix("Register consumer %s to Babylon"), opConsumerId)
 
+	// Create new embedded postgres db instance for testing verifier daemon
+	postgres := epg.NewDatabase(epg.DefaultConfig().Username("postgres").Password("postgres").Database("babylon").Port(5433))
+	err = postgres.Start()
+	require.NoError(t, err)
```
bap2pecs, line 123: can we make it optional so tests that don't need to run verifier won't need to start a db?

actually I guess we won't need to to worry about it once we changed https://github.com/babylonchain/babylon-finality-gadget/pull/62 to use https://github.com/etcd-io/bbolt?

itest/opstackl2/op_test_manager.go
```diff
@@ -109,6 +117,26 @@ func StartOpL2ConsumerManager(t *testing.T, numOfConsumerFPs uint8) *OpL2Consume
...
+
+	// create verifier daemon
```
bap2pecs, line 125: if the verifier is only used in that one test, should we move it to its own helper function and create the daemon within that function?

itest/opstackl2/op_e2e_test.go
```diff
@@ -177,3 +179,43 @@ func TestFinalityStuckAndRecover(t *testing.T) {
...
+	t.Logf(log.Prefix("Both FP instances signed the second block"))
+
+	// run the verifier daemon and process 2 blocks
+	t.Logf(log.Prefix("Starting verifier daemon"))
+	err := ctm.verifier.ProcessNBlocks(context.Background(), 2)
+	require.NoError(t, err)
+
+	// get latest finalized block via API and check response
+	fmt.Printf("targetBlockHeight: %d\n", targetBlockHeight)
+	checkLatestConsecutivelyFinalizedBlock(t, 2)
```
bap2pecs, line 220: so 2 mean latest finalized block with consecutive quorom is block at height 2? then where do we need `targetBlockHeight`? I am a bit confused

go.mod
parketh: added a replace later in the same file. There was some dependency bug when specifying the version directly here

itest/opstackl2/op_test_manager.go
parketh: yes i think this will be superseded by the db refactor. what we can do is start a db only if db configs are provided (ie make it optional)

itest/opstackl2/op_test_manager.go
parketh: i think its better to encapsulate all the different programs inside the test manager. We can add an option to disable the verifier daemon if we are concerned about overhead

itest/opstackl2/op_e2e_test.go
parketh: when run normally `ctm.verifier.ProcessBlocks()` will run indefinitely, polling for and processing new blocks at a fixd interval. However for testing we use a different function `ctm.verifier.ProcessNBlocks()` with finite execution time which terminates after processing N blocks. In L215 we specify N = 2 blocks, so expect the DB to return the latest consecutively finalized block as N = 2.

bap2pecs: I just thought of it again. I think a better way is just to use the `require.Eventually` loop to wait for `checkLatestConsecutivelyFinalizedBlock(targetBlockHeight+1)` to return true. you can see many examples in this file or the test manager

so in this way, we don't need to create a special function only for testing purpose

